Issue #125, opened 2019-09-16
Remove all BRENDA code.


SolomonShorser-OICR:
If we're not going to link to BRENDA anymore, let's remove the BRENDA-related code to make the codebase a little neater and cleaner.

PR #128, opened 2019-10-07
Feature/target pathogen

SolomonShorser-OICR:
New functionality to integrate identifiers from Target Pathogen database.

### DIFF
--- a/src/main/java/org/reactome/addlinks/AddLinks.java
+++ b/src/main/java/org/reactome/addlinks/AddLinks.java
@@ -58,6 +58,7 @@ import org.reactome.addlinks.referencecreators.ENSMappedIdentifiersReferenceCrea
 import org.reactome.addlinks.referencecreators.NCBIGeneBasedReferenceCreator;
 import org.reactome.addlinks.referencecreators.OneToOneReferenceCreator;
 import org.reactome.addlinks.referencecreators.RHEAReferenceCreator;
+import org.reactome.addlinks.referencecreators.TargetPathogenReferenceCreator;
 import org.reactome.addlinks.referencecreators.UPMappedIdentifiersReferenceCreator;
 import org.reactome.release.common.database.InstanceEditUtils;
 
@@ -510,6 +511,12 @@ public class AddLinks
 					{
 						// COSMIC should use ALL human ReferenceGeneProduct, regarless of source database.
 						sourceReferences = this.objectCache.getBySpecies("48887", "ReferenceGeneProduct");
+					}
+					else if ( refCreator instanceof TargetPathogenReferenceCreator)
+					{
+						// The list of source references is a list of reactions AND a list of EWASes.
+						sourceReferences = this.objectCache.getReactionsByID().values().stream().collect(Collectors.toList());
+						sourceReferences.addAll(this.dbAdapter.fetchInstancesByClass(ReactomeJavaConstants.EntityWithAccessionedSequence));
 					}
 					else
 					{
--- a/src/main/java/org/reactome/addlinks/AddLinks.java
+++ b/src/main/java/org/reactome/addlinks/AddLinks.java
@@ -508,10 +508,15 @@ public class AddLinks
 					else if ( refCreator instanceof TargetPathogenReferenceCreator && refCreator.getSourceRefDB().equals(REACTOME))
 					{
 						// The list of source references is a list of reactions. TargetPathogenReferenceCreator does
-						// not need special code if it's *not* creating references for reactions.
+						// not need special handling if it's *not* creating references for reactions.
+						// Creating ReferenceGeneProducts from UniProt identfiers works like other ReferenceCreators...
+						// 
+						// The reason this code is here is because the database does not contain Reactome as a 
+						// ReferenceDatabase so looking for Reactions in the cache whose source is Reactome will fail.
+						// And even if Reactome *was* a ReferenceDatabase, somewhere there would need to be code to check 
+						// the stableIdentifier attribute (via DatabaseObject) instead of the regular identifier attribute
+						// attached to the Reaction.
 						sourceReferences = this.objectCache.getReactionsByID().values().stream().collect(Collectors.toList());
-						// ... AND a list of EWASes.
-//						sourceReferences.addAll(this.dbAdapter.fetchInstancesByClass(ReactomeJavaConstants.EntityWithAccessionedSequence));
 					}
 					else
 					{
--- a/src/main/java/org/reactome/addlinks/AddLinks.java
+++ b/src/main/java/org/reactome/addlinks/AddLinks.java
@@ -26,8 +26,6 @@ import org.apache.logging.log4j.Logger;
 import org.gk.model.GKInstance;
 import org.gk.model.ReactomeJavaConstants;
 import org.gk.persistence.MySQLAdaptor;
-import org.gk.schema.InvalidAttributeException;
-import org.gk.schema.InvalidAttributeValueException;
 import org.reactome.addlinks.brenda.BRENDAReferenceDatabaseGenerator;
 import org.reactome.addlinks.dataretrieval.FileRetriever;
 import org.reactome.addlinks.dataretrieval.UniprotFileRetriever;
@@ -60,17 +58,21 @@ import org.reactome.addlinks.referencecreators.OneToOneReferenceCreator;
 import org.reactome.addlinks.referencecreators.RHEAReferenceCreator;
 import org.reactome.addlinks.referencecreators.TargetPathogenReferenceCreator;
 import org.reactome.addlinks.referencecreators.UPMappedIdentifiersReferenceCreator;
-import org.reactome.release.common.database.InstanceEditUtils;
 
 
 public class AddLinks
 {
+	private static final String PROTEIN = "PROTEIN";
+	private static final String CTD_PROCESSOR = "CTDProcessor";
+	private static final String BRENDA_UC = "BRENDA";
+	private static final String KEGG = "KEGG";
+	private static final String BRENDA = "Brenda";
+	private static final String ENSEMBL = "ENSEMBL";
+	private static final String REACTOME = "Reactome";
+	private static final String HUMAN_SPECIES_ID = "48887";
 	private static final String LINK_CHECK_REPORTS_PATH = "reports/linkCheckReports";
-
 	private static final String DUPE_REPORTS_PATH = "reports/duplicateReports";
-
 	private static final String DIFF_REPORTS_PATH = "reports/diffReports";
-
 	private static final String DATE_PATTERN_FOR_FILENAMES = "yyyy-MM-dd_HHmmss";
 
 	private static final Logger logger = LogManager.getLogger();
@@ -78,33 +80,24 @@ public class AddLinks
 	private ReferenceObjectCache objectCache;
 	
 	private List<String> fileProcessorFilter;
-	
 	private List<String> fileRetrieverFilter;
-	
 	private List<String> referenceCreatorFilter;
 	
 	private Map<String, UniprotFileRetriever> uniprotFileRetrievers;
-	
 	private Map<String, EnsemblFileRetriever> ensemblFileRetrievers;
-	
 	private Map<String, EnsemblFileRetriever> ensemblFileRetrieversNonCore;
 	
 	private Map<String, FileProcessor<?>> fileProcessors;
-	
 	private Map<String,FileRetriever> fileRetrievers;
-	
 	private Map<String, Map<String, ?>> referenceDatabasesToCreate;
-	
 	private Map<String, Object> processorCreatorLink;
 	
 	private Map<String, UPMappedIdentifiersReferenceCreator> uniprotReferenceCreators;
-	
 	private Map<String, BatchReferenceCreator<?>> referenceCreators;
 	
 	private List<String> referenceDatabasesToLinkCheck;
 	
 	private float proportionToLinkCheck = 0.1f;
-	
 	private int maxNumberLinksToCheck = 100;
 	
 	private EnsemblBatchLookup ensemblBatchLookup;
@@ -244,9 +237,9 @@ public class AddLinks
 			// LinksToCheckCache.getRefDBsToCheck() should return a list that contains everything
 			// from the Spring file AND all of the ENSEMBL and KEGG species-specific reference database names.
 			if (LinksToCheckCache.getRefDBsToCheck().contains(refDBInst.getDisplayName())
-					|| (refDBInst.getDisplayName().toUpperCase().contains("ENSEMBL") && LinksToCheckCache.getRefDBsToCheck().contains("ENSEMBL"))
-					|| (refDBInst.getDisplayName().toUpperCase().contains("BRENDA") && LinksToCheckCache.getRefDBsToCheck().contains("Brenda"))
-					|| (refDBInst.getDisplayName().toUpperCase().contains("KEGG") && LinksToCheckCache.getRefDBsToCheck().contains("KEGG"))
+					|| (refDBInst.getDisplayName().toUpperCase().contains(ENSEMBL) && LinksToCheckCache.getRefDBsToCheck().contains(ENSEMBL))
+					|| (refDBInst.getDisplayName().toUpperCase().contains(BRENDA_UC) && LinksToCheckCache.getRefDBsToCheck().contains(BRENDA))
+					|| (refDBInst.getDisplayName().toUpperCase().contains(KEGG) && LinksToCheckCache.getRefDBsToCheck().contains(KEGG))
 				)
 			{
 				if (LinksToCheckCache.getCache().get(refDBInst).size() > 0)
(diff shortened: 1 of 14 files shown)


src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -58,52 +56,48 @@
...
+import org.reactome.addlinks.referencecreators.TargetPathogenReferenceCreator;
 import org.reactome.addlinks.referencecreators.UPMappedIdentifiersReferenceCreator;
-import org.reactome.release.common.database.InstanceEditUtils;
 
 
 public class AddLinks
 {
+	private static final String PROTEIN = "PROTEIN";
+	private static final String CTD_PROCESSOR = "CTDProcessor";
+	private static final String BRENDA_UC = "BRENDA";
```
jweiser: Is BRENDA still needed?

SolomonShorser-OICR: There is a separate issue for removing all BRENDA-related code: https://github.com/reactome/add-links/issues/125

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
```diff
@@ -0,0 +1,219 @@
...
+						{
+							if (!reactomeToTargetPathogen.containsKey(identifier))
+							{
+								reactomeToTargetPathogen.put(identifier, new HashSet<>(Arrays.asList(targetPathogenIdentifier)));
+							}
+							else
+							{
+								Set<String> tmpSet = reactomeToTargetPathogen.get(identifier);
+								tmpSet.add(targetPathogenIdentifier);
+								reactomeToTargetPathogen.put(identifier, tmpSet);
```
jweiser: Could this be written as:
```
reactomeToTargetPathogen.get(identifier).add(targetPathogenIdentifier);
```

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
```diff
@@ -0,0 +1,219 @@
...
+						{
+							if (!uniProtToTargetPathogen.containsKey(identifier))
+							{
+								uniProtToTargetPathogen.put(identifier, new HashSet<>(Arrays.asList(targetPathogenIdentifier)));
+							}
+							else
+							{
+								Set<String> tmpSet = uniProtToTargetPathogen.get(identifier);
+								tmpSet.add(targetPathogenIdentifier);
+								uniProtToTargetPathogen.put(identifier, tmpSet);
```
jweiser: Same as above?

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
SolomonShorser-OICR: It might be possible to replace the first two lines with one line, but I'm not sure what the benefit is. I sometimes find such expressions can be harder to debug when NPEs or other issues pop up.

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
SolomonShorser-OICR: https://github.com/reactome/add-links/pull/128/files#r332185224

src/test/java/org/reactome/addlinks/test/TestTargetPathogenFileProcessor.java
```diff
@@ -0,0 +1,54 @@
...
+	private Path pathToFile;
+
+	private final static String fileContent = "http://target.sbg.qb.fcen.uba.ar/patho/protein/56425908be737e6c7a9fd4d0,H37Rv,Rv0046c|ino1,R-MTU-879331,P9WKI1\n" + 
+			"http://target.sbg.qb.fcen.uba.ar/patho/protein/56425908be737e6c7a9fd525,H37Rv,Rv0126|treS,R-MTU-868709,P9WQ19\n" + 
+			"http://target.sbg.qb.fcen.uba.ar/patho/protein/5b2801b3be737e35a6df1a69,MeloI,Minc3s11049g44572,R-CEL-74723,P34462\n" + 
+			"http://target.sbg.qb.fcen.uba.ar/patho/protein/5b2801b3be737e35a6df151a,MeloI,Minc3s09360g43141|RP-L40e,|RPL40,R-XTR-983157,F6VNK4\n" + 
+			"http://target.sbg.qb.fcen.uba.ar/patho/protein/5b2801b3be737e35a6df151a,MeloI,Minc3s09360g43141|RP-L40e,|RPL40,R-CFA-983157,F1PDG4\n" + 
+			"http://target.sbg.qb.fcen.uba.ar/patho/protein/5b354852be737e166372e8dd,EcoAIEC_C7225,AIEC_C7225_05581|APU18_05390| pNDM102337_153|groEL| MS6198_A148| CA268_28940| pNDM10505_155|BET08_16250| NDM1Dok01_N0169,R-HSA-8850539,P48643\n" + 
+			"http://target.sbg.qb.fcen.uba.ar/patho/protein/5b354852be737e166372e8dd,EcoAIEC_C7225,AIEC_C7225_05581|APU18_05390| pNDM102337_153|groEL| MS6198_A148| CA268_28940| pNDM10505_155|BET08_16250| NDM1Dok01_N0169,R-SSC-8850539,I3LR32\n";
```
jweiser: Could the "http://target.sbg.qb.fcen.uba/ar/patho/protein/" part of the URL be extracted into a constant so it's not repeated?

It also may be clearer to use String.join("\n", ...) instead of concatenating all the Strings with a new-line at the end of each.

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
jweiser: I find it easier to understand what is going on without the creation and re-assignment of the temporary set.

For this logic, I don't think a NullPointerException should ever happen because this happens after a check that the identifier does exist and so the `reactomeToTargetPathogen.get(identifier)` should always return a set that you can add to.

src/test/java/org/reactome/addlinks/test/TestTargetPathogenFileProcessor.java
SolomonShorser-OICR: Ok, I'll refactor the URL.

I usually avoid using `String.join` for situations that involve string literals.

jweiser: Okay, that's fine.

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
SolomonShorser-OICR: Personally, I like the explicitness of it. And sure, an NPE probably should never happen, but I've been trying to get into this habit because in other situations, too many chained calls has resulted in having to split everything up to trace exactly where the error was. So now I'm trying to get _out_ of the habit of too many chained function calls.

jweiser: Okay, I'll leave it up to you.  

I agree that multiple chained calls can be difficult to debug.  I was just thinking in this case it would be okay for readability since it is only one chained call and the first method call is guaranteed to return a non-null value.

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -58,52 +56,48 @@
...
+	private static final String KEGG = "KEGG";
+	private static final String BRENDA = "Brenda";
+	private static final String ENSEMBL = "ENSEMBL";
+	private static final String REACTOME = "Reactome";
+	private static final String HUMAN_SPECIES_ID = "48887";
```
cookersjs: I don't quite know how AddLinks accesses these values, whether they're only used in this file or elsewhere, but depending on if this list grows maybe we could consider an AddLinksConstants class at some point in the future?

src/main/java/org/reactome/addlinks/fileprocessors/TargetPathogenFileProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+{
+
+	public TargetPathogenFileProcessor(String processorName)
+	{
+		super(processorName);
+	}
+
+	public TargetPathogenFileProcessor()
+	{
+		super(null);
```
cookersjs: I'm a bit confused by the initiation of the class with `super(null)` here, with an additional super above that can be invoked

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
```diff
@@ -0,0 +1,219 @@
...
+
+	/**
+	 * Can create references to Target Pathogen on reaction objects that are identifier by a Reactome identifier. We will also probably want to create a reference
+	 * from proteins identified by UniProt identifiers. Might be tricky to combine this functionality into one ReferenceCreator, but we'll see...
+	 */
+	@Override
+	public void createIdentifiers(long personID, Map<String, String> mapping, List<GKInstance> sourceReferences) throws Exception
+	{
+		this.personID = personID;
+		if (reactomeToTargetPathogen.size() == 0 && uniProtToTargetPathogen.size() == 0)
```
cookersjs: Could it ever happen that one is populated but the other is not?

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: If it makes sense. Most constants are declared in the context in which they are used. If it looks like there are many constants being used across multiple classes, this is something that could be considered.

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
SolomonShorser-OICR: If both sets are empty, the code between lines 53 and 86 is supposed to process ALL identifiers and populate both caches. At the end of that loop, both sets ideally should have some data. If one of them somehow does _not_ have data, then it probably means there were no identifiers for it, and trying to repopulate it separately wouldn't do much because the input will be the same as the previous attempt.

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
```diff
@@ -0,0 +1,219 @@
...
+			this.logger.error("sourceRefDB was: {} and classReferring was: {} - THIS IS NOT A VALID COMBINATION! Valid options are: UniProt/ReferenceGeneProduct and Reactome/Reaction.", this.sourceRefDB, this.classReferringToRefName);
+		}
+	}
+
+	/**
+	 * Create references for Reactome Identifiers.
+	 * @param sourceReferences
+	 * @throws Exception
+	 */
+	private void createReferencesForReactomeIdentifiers(List<GKInstance> sourceReferences) throws Exception
```
jweiser: This method and `createReferencesForUniProtIdentifiers` look almost identical.  Can the common logic be extracted out to remove the duplication?

SolomonShorser-OICR: They differ in terms of which static-member cache they look things up in, as well as what value they use as as lookup key. They look similar, but there are subtle differences, and I'm not sure they are easy to refactor away, or that it's even worth the extra effort. Still, I'll give it another look...

jweiser: I was thinking something like this (starting from line 100 in the file):
```
	// So... which instance are we? We can determine from how the SourceRefDB and classReferring are set.
	if (this.sourceRefDB.equals("UniProt") && this.classReferringToRefName.equals(ReactomeJavaConstants.ReferenceGeneProduct))
	{
		this.createReferencesForIdentifiers(
			filterGKInstancesByClass(sourceReferences, ReactomeJavaConstants.ReferenceGeneProduct),
			this::getTargetPathogensFromUniProtSourceRef
		);
	}
	else if (this.sourceRefDB.equals("Reactome") && this.classReferringToRefName.equals(ReactomeJavaConstants.Reaction))
	{
		this.createReferencesForIdentifiers(
			filterGKInstancesByClass(sourceReferences, ReactomeJavaConstants.Reaction),
			this::getTargetPathogensFromReactomeStableIdentifierSourceRef
		);
	}
	else
	{
		this.logger.error("sourceRefDB was: {} and classReferring was: {} - THIS IS NOT A VALID COMBINATION! Valid options are: UniProt/ReferenceGeneProduct and Reactome/Reaction.", this.sourceRefDB, this.classReferringToRefName);
	}
}

private void createReferencesForIdentifiers(
	List<GKInstance> sourceReferences,
	Function<GKInstance, Set<String>> sourceRefToTargetPathogens
) throws Exception {
	int numRefsCreated = 0;
	int numRefsPreexisting = 0;
	int numTargetIdentifiers = 0;
	int[] counts = {numTargetIdentifiers, numRefsCreated, numRefsPreexisting};

	for (GKInstance sourceRef : sourceReferences) {
		for (String targetPathogenID : sourceRefToTargetPathogens.apply(sourceRef)) {
			counts = createReferencesForIdentifier(sourceRef, targetPathogenID, counts);
		}
	}

	this.logger.info("Inspected {} objects, {} identifiers for TargetPathogen.\n"
	+ " {} new references were created.\n"
	+ "{} references already existed (and were NOT re-created).",
	sourceReferences.size(), numTargetIdentifiers, numRefsCreated, numRefsPreexisting);
}

private List<GKInstance> filterGKInstancesByClass(List<GKInstance> gkInstances, String className) {
	return gkInstances.stream().filter(
		sr -> sr.getSchemClass().getName().equals(className)
	).collect(Collectors.toList());
}

private Set<String> getTargetPathogensFromUniProtSourceRef(GKInstance sourceRef) {
	Set<String> targetPathogens = new HashSet<>();
	try {
		String uniprotIdentifier = sourceRef.getAttributeValue(ReactomeJavaConstants.identifier).toString();
		targetPathogens = uniProtToTargetPathogen.computeIfAbsent(uniprotIdentifier, k -> new HashSet<>());
	} catch (Exception e) {
		this.logger.error("Unable to get UniProtIdentifier from {}", sourceRef.getExtendedDisplayName(), e);
	}

	return targetPathogens;
}

private Set<String> getTargetPathogensFromReactomeStableIdentifierSourceRef(GKInstance sourceRef) {
	Set<String> targetPathogens = new HashSet<>();

	try {
		GKInstance reactomeStableIdentifier = (GKInstance) sourceRef.getAttributeValue(ReactomeJavaConstants.stableIdentifier);
		String identifier = reactomeStableIdentifier.getAttributeValue(ReactomeJavaConstants.identifier).toString();
		targetPathogens = reactomeToTargetPathogen.computeIfAbsent(identifier, k -> new HashSet<>());
	} catch (Exception e) {
		this.logger.error("Unable to get Reactome Stable Identifier from {}", sourceRef.getExtendedDisplayName(), e);
	}

	return targetPathogens;
}
```

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
jweiser: I realized you could do this:
`reactomeToTargetPathogen.computeIfAbsent(identifier, k -> new HashSet()).add(targetPathogenIdentifier);`

The "computeIfAbsent" will guarantee a key that isn't found will create an empty set to be added to and guarantee avoidance of a NPE

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
jweiser: https://github.com/reactome/add-links/pull/128#discussion_r334081844?

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
SolomonShorser-OICR: I still don't like the chained `.add` here. And I'm not sure I like `computeIfAbsent` either. In this context, it seems to simply hide a very simple logical expression. I'm not sure I see the value of it in this situation.

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
SolomonShorser-OICR: Hmmm... interesting, but what problem does this large rewrite solve?

jweiser: Some of the logic that was duplicated between related methods gets consolidated.

SolomonShorser-OICR: I still don't think the duplicate was _that_ big a deal. Still, I'll look into implementing something like this and then testing, if it's going to be a blocker. But this is a (relatively) large-ish rewrite, it will need to wait...

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
jweiser: Okay, no problem if you prefer the current implementation.  It's just my preference not to reassign to a map if a value already exists.

src/main/java/org/reactome/addlinks/referencecreators/TargetPathogenReferenceCreator.java
jweiser: Removing duplication is something I try to be pretty aggressive about.  It makes it easier if changes are needed so only one area has to be modified and logic doesn't get out of sync.

I try to stick as closely as I can to the DRY principle using simple methods so things are easier to unit test and reason about.

SolomonShorser-OICR:
> Per our discussion, refactoring similar to [#128 (comment)](https://github.com/reactome/add-links/pull/128#discussion_r334080201) would be great.

Yup. I'll put it in my list of things to get to...

Issue #130, opened 2019-11-18
Wrong species code for some KEGG identifiers


SolomonShorser-OICR:
In R71, link was created for identifier mtv:Rv2540c but it should have been mtu:Rv2540c. Possible that there are conflicts in species codes for KEGG.

SolomonShorser-OICR:
It looks like the programmatic list of KEGG species doesn't make an easy distinction between mtu and mtv. See: http://www.genome.jp/kegg-bin/download_htext?htext=br08601.keg&format=htext&filedir= and also https://www.genome.jp/kegg/catalog/org_list.html

In the first link, "Mycobacterium tuberculosis H37Rv" appears _twice_, once with the mtu prefix and once with the mtv prefix. Nothing else distinguishes these, and because the species names are identical, it means that performing a lookup by species name could cause problems.

In the second link, it can be seen that mtu refers to data coming from RefSeq and mtv refers to data whose source is GeneBank.

Might need to add special code for M. Tuberculosis to include the prefix in the species name, so: "Mycobacterium tuberculosis H37Rv (mtu)" and "Mycobacterium tuberculosis H37Rv (mtv)". It will make look-ups by species-name more complicated, so maybe look-ups should be prefix-match rather then requiring a full-string match...

PR #133, opened 2020-01-02
Feature/kegg mtu vs mtv

SolomonShorser-OICR:
For mycobacterium tuberculosis, append the KEGG prefix to the ReferenceDatabase name. Also introduce the concept of "forbidden" KEGG prefixes - identifiers that have such a prefix will not be added to the database.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,7 +52,7 @@
 		<dependency>
 			<groupId>org.apache.logging.log4j</groupId>
 			<artifactId>log4j-core</artifactId>
-			<version>2.8.2</version>
+			<version>2.13.0</version>
 		</dependency>
 		<dependency>
 			<groupId>commons-net</groupId>
@@ -162,7 +162,7 @@
 		<dependency>
 			<groupId>org.reactome.release</groupId>
 			<artifactId>release-common-lib</artifactId>
-			<version>1.0.2</version>
+			<version>1.2.0</version>
 		</dependency>
 	</dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -155,10 +155,24 @@
 			<version>1.6</version>
 		</dependency>
 		<dependency>
+			<!--
+			***NOTE:***
+			For this dependency, you must MANUALLY install jnlp-1.2 from here:
+			https://github.com/reactome/CuratorTool/raw/master/lib/jnlp-1.2.jar
+			If you don't do this, you won't be able to build.
+			-->
 			<groupId>org.reactome.release</groupId>
 			<artifactId>release-common-lib</artifactId>
 			<version>1.2.0</version>
+			<exclusions>
+				<exclusion>
+					<artifactId>jnlp-api</artifactId>
+					<groupId>javax.jnlp</groupId>
+				</exclusion>
+			</exclusions>
 		</dependency>
+
+
 	</dependencies>
 
 	<build>
@@ -264,59 +278,4 @@
 			</resource>
 		</resources>
 	</build>
-
-	<repositories>
-		<repository>
-			<id>nexus-ebi-repo</id>
-			<name>The EBI internal repository</name>
-			<url>http://www.ebi.ac.uk/Tools/maven/repos/content/groups/ebi-repo/</url>
-			<releases>
-				<enabled>true</enabled>
-			</releases>
-			<snapshots>
-				<enabled>false</enabled>
-			</snapshots>
-		</repository>
-		<!-- EBI SNAPSHOT repo -->
-		<repository>
-			<id>nexus-ebi-snapshot-repo</id>
-			<name>The EBI internal snapshot repository</name>
-			<url>http://www.ebi.ac.uk/Tools/maven/repos/content/groups/ebi-snapshots/</url>
-			<releases>
-				<enabled>false</enabled>
-			</releases>
-			<snapshots>
-				<enabled>true</enabled>
-			</snapshots>
-		</repository>
-		<repository>
-			<name>EBI</name>
-			<url>http://www.ebi.ac.uk/Tools/maven/repos/content/repositories/releases/</url>
-			<id>EBI</id>
-			<releases>
-				<enabled>true</enabled>
-			</releases>
-		</repository>
-		<!-- OICR respository for Reactome artefacts -->
-		<repository>
-			<id>oicr-central</id>
-			<name>artifacts.oicr.on.ca-releases</name>
-			<url>https://artifacts.oicr.on.ca/artifactory/reactome</url>
-			<releases>
-				<enabled>true</enabled>
-			</releases>
-		</repository>
-		<repository>
-			<id>oicr-snapshots</id>
-			<name>artifacts.oicr.on.ca-snapshots</name>
-			<url>https://artifacts.oicr.on.ca/artifactory/reactome</url>
-			<releases>
-				<enabled>false</enabled>
-			</releases>
-			<snapshots>
-				<enabled>true</enabled>
-			</snapshots>
-		</repository>
-	</repositories>
-
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -100,12 +100,12 @@
 		<dependency>
 			<groupId>commons-discovery</groupId>
 			<artifactId>commons-discovery</artifactId>
-			<version>0.4</version>
+			<version>0.5</version>
 		</dependency>
 		<dependency>
 			<groupId>mysql</groupId>
 			<artifactId>mysql-connector-java</artifactId>
-			<version>5.1.40</version>
+			<version>5.1.48</version>
 		</dependency>
 		<dependency>
 			<groupId>org.powermock</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -59,11 +59,6 @@
 			<artifactId>commons-net</artifactId>
 			<version>3.5</version>
 		</dependency>
-		<!-- <dependency>
-			<groupId>org.reactome.server</groupId>
-			<artifactId>reactome-core</artifactId>
-			<version>1.1.2</version>
-		</dependency> -->
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-beans</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -157,19 +157,21 @@
 		<dependency>
 			<!--
 			***NOTE:***
-			For this dependency, you must MANUALLY install jnlp-1.2 from here:
+			For this dependency, you MUST *manually* install jnlp-1.2 from here:
 			https://github.com/reactome/CuratorTool/raw/master/lib/jnlp-1.2.jar
 			If you don't do this, you won't be able to build.
 			-->
 			<groupId>org.reactome.release</groupId>
 			<artifactId>release-common-lib</artifactId>
 			<version>1.2.0</version>
-			<exclusions>
+			<!-- I tried to exclude jnlp but it doesn't work because jnlp is
+			a dependency of reactome-base, not release-common-lib. -->
+<!-- 			<exclusions>
 				<exclusion>
 					<artifactId>jnlp-api</artifactId>
 					<groupId>javax.jnlp</groupId>
 				</exclusion>
-			</exclusions>
+			</exclusions> -->
 		</dependency>
 
 
--- a/pom.xml
+++ b/pom.xml
@@ -59,11 +59,11 @@
 			<artifactId>commons-net</artifactId>
 			<version>3.5</version>
 		</dependency>
-		<dependency>
+		<!-- <dependency>
 			<groupId>org.reactome.server</groupId>
 			<artifactId>reactome-core</artifactId>
 			<version>1.1.2</version>
-		</dependency>
+		</dependency> -->
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-beans</artifactId>
--- a/src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
+++ b/src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
@@ -23,22 +23,23 @@ public class KEGGReferenceDatabaseGenerator
 	private static ReferenceDatabaseCreator dbCreator;
 	private static MySQLAdaptor adaptor;
 	private static Map<String, GKInstance> keggCodesToRefDBMap = new HashMap<String, GKInstance>();
-	
+	private static final String KEGG_GENE = "KEGG Gene";
+
 	// private constructor to avoid instantiation.
 	private KEGGReferenceDatabaseGenerator()
 	{
 	}
-	
+
 	public static void setDBAdaptor(MySQLAdaptor adaptor)
 	{
 		KEGGReferenceDatabaseGenerator.adaptor = adaptor;
 	}
-	
+
 	public static void setDBCreator(ReferenceDatabaseCreator creator)
 	{
 		KEGGReferenceDatabaseGenerator.dbCreator = creator;
 	}
-	
+
 	private static Long createReferenceDatabase(String dbName, String speciesName, String speciesURL, ReferenceObjectCache objectCache) throws Exception
 	{
 		if (objectCache.getRefDbNamesToIds().keySet().contains(dbName))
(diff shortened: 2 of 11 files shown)


src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
```diff
@@ -194,43 +200,48 @@ public static String generateDBNameFromReactomeSpecies(ReferenceObjectCache obje
...
+	 * is a different between the KEGG and Reactome species names. Use this method to create a new ReferenceDatabase object based on the KEGG
 	 * prefix and species name. Rebuilding caches is <em>STRONGLY</em> recommended after calling this method.
 	 * @param keggPrefix
 	 * @param keggSpeciesName
 	 * @return The name of the new ReferenceDatabase.
 	 */
 	public static synchronized String createReferenceDatabaseFromKEGGData(String keggPrefix, String keggSpeciesName, ReferenceObjectCache objectCache)
 	{
+		String newDBName = KEGG_GENE + " ("+keggSpeciesName + ")";
+		if (keggSpeciesName.contains("Mycobacterium tuberculosis H37Rv")) // if we start needing to handle more exceptional cases than just Mycobacterium tuberculosis H37Rv, we may need to rethink the general design...
```
cookersjs: I wonder when other species in KEGG will also be subdivided into different strains

SolomonShorser-OICR: Other species of single-cell organisms probably _are_. But so far, we have only had to deal with this one.

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreator.java
```diff
@@ -95,48 +95,51 @@ public void createIdentifiers(long personID, Map<String, List<Map<KEGGKeys, Stri
...
+							{
+								names.add(keggIdentifier);
+							}
+							Map<String,List<String>> extraAttributes = new HashMap<>(1);
+							extraAttributes.put(ReactomeJavaConstants.name, names);
+							logger.trace("For {}, creating new KEGG xref: {}",sourceReference.getDisplayName(), keggIdentifier);
+							if (!this.testMode)
 							{
-								this.refCreator.createIdentifier(keggGeneIdentifier, String.valueOf(sourceReference.getDBID()),targetDB, personID, this.getClass().getName(), speciesID, extraAttributes);
+								if (!this.testMode && targetDB != null)
```
cookersjs: Unnecessary condition (`!this.testMode`) since it is checked right above

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreatorHelper.java
```diff
@@ -57,22 +62,23 @@ private synchronized String createNewKEGGReferenceDatabase(String targetIdentifi
...
+		if (identifier.startsWith("vg:"))
 		{
 			targetDB = "KEGG Gene (Viruses)";
-			keggIdentifier = keggIdentifier.replaceFirst("vg:", "");
+			identifier = identifier.replaceFirst("vg:", "");
 		}
-		else if (keggIdentifier.startsWith("ag:"))
+		else if (identifier.startsWith("ag:"))
 		{
 			targetDB = "KEGG Gene (Addendum)";
```
cookersjs: I know this isn't a new thing, but I'm curious about where this DB comes from. Does Addendum imply it is like a miscellaneous Kegg database? I tried searching for it on KEGG but nothing came up

SolomonShorser-OICR: On this page: https://www.genome.jp/kegg/genes.html , the first table contains "

"Addendum    ag    PubMed-based collection of functionally characterized proteins"

You can search this specific database here: https://www.genome.jp/dbget-bin/www_bfind?ag

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreator.java
```diff
@@ -95,48 +95,51 @@ public void createIdentifiers(long personID, Map<String, List<Map<KEGGKeys, Stri
...
+							keggIdentifier = keggGeneIdentifier;
 						}
-						Map<String,List<String>> extraAttributes = new HashMap<String,List<String>>(1);
-						extraAttributes.put(ReactomeJavaConstants.name, names);
-						logger.trace("For {}, creating new KEGG xref: {}",sourceReference.getDisplayName(), keggIdentifier);
-						if (!this.testMode)
+
+						logger.trace("Working on source object: {}", sourceReference.toString());
+						String targetDB = null;
+						KEGGReferenceCreatorHelper referenceCreatorHelper = new KEGGReferenceCreatorHelper(objectCache, logger);
```
jweiser: Should `logger` be referenced as `this.logger` here and on line 107 for consistency with the rest of the codebase?

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreator.java
SolomonShorser-OICR: Fixed.

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreatorHelper.java
```diff
@@ -81,18 +87,28 @@ else if (keggIdentifier.startsWith("ag:"))
 			{
 				targetDB = targetDBID.toString();
 			}
-			// If targetDB is STILL NULL, it means we weren't able to determine which KEGG ReferenceDatabase to use for this keggIdentifier. 
-			// So, we can't add the cross-reference since we don't know which species-specific ReferenceDatabase to use. 
+			// If targetDB is STILL NULL, it means we weren't able to determine which KEGG ReferenceDatabase to use for this keggIdentifier.
+			// So, we can't add the cross-reference since we don't know which species-specific ReferenceDatabase to use.
 			if (targetDB == null)
```
jweiser: This check shouldn't be necessary.  If the `else` block is entered, it's guaranteed that `targetDB` will still be null as it was set on line 70.

pom.xml
```diff
@@ -162,7 +157,7 @@
 		<dependency>
 			<groupId>org.reactome.release</groupId>
 			<artifactId>release-common-lib</artifactId>
-			<version>1.0.2</version>
+			<version>1.2.0</version>
```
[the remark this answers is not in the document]
SolomonShorser-OICR: I'll test building without the `<repositories>` section.

jweiser:
> For mycobacterium tuberculosis, append the KEGG prefix to the ReferenceDatabase name. Also introduce the concept of "forbidden" KEGG prefixes - identifiers that have such a prefix will not be added to the database.

So just to confirm, we've decided "mtv" is forbidden to be added to the database?

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreator.java
SolomonShorser-OICR: Hmm... possibly...

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreatorHelper.java
SolomonShorser-OICR: Might be a legacy from older logic. Will look more closely at this...

SolomonShorser-OICR:
> So just to confirm, we've decided "mtv" is forbidden to be added to the database?

Peter and Bijay had a preference for keeping mtu, so mtv will not be used.

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreatorHelper.java
SolomonShorser-OICR: Wait a minute... what are you looking at? There is no `else` block...

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreator.java
SolomonShorser-OICR: It's consistent within the class. Earlier, I fell out of the habit of qualifying loggers with `this`.

SolomonShorser-OICR: Yes, the codebase is not 100% consistent, but I'm more focused on fixing the bug here, then making everything consistent.

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreatorHelper.java
jweiser: The else block starting on line 83.  It contains this line within it.

jweiser:
> > So just to confirm, we've decided "mtv" is forbidden to be added to the database?
> 
> Peter and Bijay had a preference for keeping mtu, so mtv will not be used.

Okay, just wanted to check :).

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreator.java
jweiser: Okay, no problem.  I just noticed the difference compared to other parts of the PR.

src/main/java/org/reactome/addlinks/referencecreators/KEGGReferenceCreatorHelper.java
SolomonShorser-OICR: Oh _that_ `else` block...

When then `else` block is entered, `targetDB` will be null. It _could_ get set on line 88 (as a result of the lookup on line 85). But if the kegg prefix does not exist in the database then `targetDB` will still be null _after_ line 88, in which case a new reference database needs to be created.

jweiser: Ah okay - line 88 must have been collapsed when I looked at it so I didn't see it could get set there.  Nevermind about this, then :).

src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
```diff
@@ -110,12 +116,12 @@ public static void generateSpeciesSpecificReferenceDatabases(ReferenceObjectCach
...
-		String newDBName = "KEGG Gene (Viruses)";
+		String newDBName = KEGG_GENE + " (Viruses)";
 		try
 		{
 			createReferenceDatabase(newDBName, "Viruses", speciesURL, objectCache);
 			speciesURL = KEGG_URL.replace("###SP3###", "ag:");
-			newDBName = "KEGG Gene (Addendum)";
+			newDBName = KEGG_GENE + " (Addendum)";
 			createReferenceDatabase(newDBName, "Addendum", speciesURL, objectCache);
 		}
```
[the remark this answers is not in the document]
SolomonShorser-OICR: The `replace` function does not modify the string it is called on - it _returns_ the result of the replacement operation. So `KEGG_URL` does not get modified by calling replace (Strings, in Java, are immutable - https://docs.oracle.com/javase/tutorial/java/data/strings.html).

`newDBName` and `speciesURL` get new values only _after_ the first call to `createReferenceDatabase` (line 122) - at that point, the name and URL for "viruses" should be in the database so it's OK to reuse the variables with new values, and then put _those_ values into the database (line 125).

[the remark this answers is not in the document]
jweiser: Yup, you're right.

Okay, so since two reference databases are created with two distinct names and URL values, should the link checking on line 132 happen twice as well?  Right now it looks like it's only called for the "Addendum" db name and not "Viruses" db name since `newDBName` gets reassigned on line 124 before the link checking happens.

pom.xml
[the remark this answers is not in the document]
SolomonShorser-OICR: I started with a clean `.m2` directory and then I got this error:
```
[ERROR] Failed to execute goal on project AddLinks: Could not resolve dependencies for project org.reactome.release:AddLinks:jar:1.3.0: Failed to collect dependencies at org.reactome.release:release-common-lib:jar:1.2.0 -> org.reactome.base:reactome-base:jar:1.2.22 -> javax.jnlp:jnlp:jar:1.2: Failed to read artifact descriptor for javax.jnlp:jnlp:jar:1.2: Could not transfer artifact javax.jnlp:jnlp:pom:1.2 from/to freehep.repository (http://java.freehep.org/maven2/): Failed to transfer file: http://java.freehep.org/maven2/javax/jnlp/jnlp/1.2/jnlp-1.2.pom. Return code is: 502 , ReasonPhrase:Bad Gateway. -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
```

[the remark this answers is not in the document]
SolomonShorser-OICR: I'm guessing the requirement for jnlp (read more on jnlp here: https://docs.oracle.com/javase/tutorial/deployment/deploymentInDepth/jnlp.html) comes from CuratorTool. I really don't think this requirement should be a "base" requirement. Maybe detangling this from reactome-base would be a good issue to work on next?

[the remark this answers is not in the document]
jweiser: Hmm, for some reason the pom.xml of 'reactome-base' and 'release-common-lib' aren't showing up on mvnrepository.com.  jnlp is listed as a dependency in the 'reactome-base' v1.2.22 pom.xml.  I'll have to look into it more.

https://mvnrepository.com/artifact/org.reactome.base/reactome-base/1.2.22
https://mvnrepository.com/artifact/org.reactome.release/release-common-lib/1.2.0

https://repo1.maven.org/maven2/org/reactome/base/reactome-base/1.2.22/reactome-base-1.2.22.pom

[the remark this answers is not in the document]
jweiser: Thanks for testing!  It's good we know about this issue with the published Maven stuff earlier on!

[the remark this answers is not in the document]
SolomonShorser-OICR: Issue created: https://github.com/reactome/CuratorTool/issues/27

src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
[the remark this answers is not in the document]
SolomonShorser-OICR: Good catch. That's been updated.

src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
```diff
@@ -90,7 +91,12 @@ public static void generateSpeciesSpecificReferenceDatabases(ReferenceObjectCach
 					try
 					{
 						String speciesURL = KEGG_URL.replace("###SP3###", code + ":");
-						String newDBName = "KEGG Gene ("+speciesName + ")";
+						String newDBName = KEGG_GENE + " (" + speciesName + ")";
+						if (code.equals("mtu") || code.equals("mtv"))
+						{
+							newDBName = KEGG_GENE + " ("+speciesName + " - " + code + ")";
+						}
+
```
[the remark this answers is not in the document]
SolomonShorser-OICR: I _could_ do this, but I am not sure I see a point in making this change.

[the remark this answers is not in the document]
jweiser: Not a significant advantage.  Just to reduce duplication of the same Strings.  Since the code is just an additional suffix, adding it instead of reassignment seemed like a good way to go.

I just found the assignment followed by a possible immediate reassignment on line 97 a bit unusual.  Another way to go could be one assignment in an if block and the other assignment in an else block.  I just find that easier to follow rather than an overwrite of an existing value.

Like I mentioned, it's not a blocker.  I'll leave it up to you.

SolomonShorser-OICR:
Fixes #130

Issue #135, opened 2020-03-04
Make link-checking more robust


SolomonShorser-OICR:
Currently, link-checking is a phase that happens after reference creation. It _only_ checks links for references that were added _during_ the release process. If the process crashes during reference creation but before link-checking begins, then when the process is re-run, all previously added references are no longer "new" and will not be eligible for link-checking. 

Options:

 1. Make link-checking happen as soon as reference creation finishes. As soon as a reference creator finishes creating references, link-checking should begin _immediately_ for the references that were created, rather than waiting until after _all_ link-checking is complete.
 2. If the link-checker cannot find any reference to check for a ReferenceDatabase it should query for the most recent _n_ references already in the database and test those.

Option #1 is favoured as it doesn't require any extra queries.

PR #136, opened 2020-03-13
Feature/improved linkchecking

SolomonShorser-OICR:
This PR deals with #135 

Link-checking will now be performed after each ReferenceCreator finishes creating references. This way, if anything interrupts AddLinks, there will at least be a partial link-checking report for the ReferenceCreators that executed prior to the interruption.


src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -193,96 +197,86 @@ public void doAddLinks() throws Exception
...
+			reportLine.append(refDBInst.getDisplayName()).append("\t");
+		}
+		else
+		{
+			try
+			{
+				Optional<String> longerName = ((List<String>)refDBInst.getAttributeValuesList(ReactomeJavaConstants.name)).stream().filter(name -> name.length() > ENSEMBL.length()).sorted().findFirst();
+				reportLine.append(longerName.orElse(ENSEMBL)).append("\t");
+			}
+			catch (InvalidAttributeException e)
```
cookersjs: How come you specifically catch the IAE? Since your just printing the stacktrace wouldn't it be ok with just the single catch?

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -193,96 +197,86 @@ public void doAddLinks() throws Exception
...
-			// from the Spring file AND all of the ENSEMBL and KEGG species-specific reference database names.
-			if (LinksToCheckCache.getRefDBsToCheck().contains(refDBInst.getDisplayName())
-					|| (refDBInst.getDisplayName().toUpperCase().contains(ENSEMBL) && LinksToCheckCache.getRefDBsToCheck().contains(ENSEMBL))
-					|| (refDBInst.getDisplayName().toUpperCase().contains(KEGG) && LinksToCheckCache.getRefDBsToCheck().contains(KEGG))
-				)
+			LinkCheckInfo result = entry.getValue();
+			int statusCode = result.getStatusCode();
+			String identifier = result.getIdentifier();
+			// If the keyword is NOT found...
+			if (!result.isKeywordFound())
```
cookersjs: In the future maybe a contingency on results that don't have a keyword could just be the presence of a '200' response? Resources like VGNC won't return a result with the keyword contained since it actually retrieves the data once you visit the URL. Would it make sense to have something along the lines of 'keyword was not found but a 200 status was received'? I know that given the way some bioinformatics resources are made, this will actually allow some incorrect ones to squeak through, but curious about if this is something we should consider

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -416,6 +410,8 @@ private void purgeUnusedRefDBs()
 	@SuppressWarnings("unchecked")
 	private void createReferences(long personID, Map<String, Map<String, ?>> dbMappings) throws IOException, Exception
 	{
+		String linkCheckReportName = LINK_CHECK_REPORTS_PATH + "/linkCheckSummaryReport" + DateTimeFormatter.ofPattern(DATE_PATTERN_FOR_FILENAMES).format(LocalDateTime.now()) + ".tsv";
+		Files.write(Paths.get(linkCheckReportName), "RefDBName\tNumOK\tNumNotOK\n".getBytes(), StandardOpenOption.CREATE);
```
cookersjs: Should the link check report have the URL in it? Or if that is adding too much text should we have the URL template somewhere?

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -514,20 +510,82 @@ else if (fileProcessorName.get() instanceof List<?>)
...
+	 */
+	private String checkLinksForRefCreator(BatchReferenceCreator<?> refCreator)
+	{
+		String targetRefDB = refCreator.getTargetRefDB();
+		String line = "";
+		if (!targetRefDB.toUpperCase().contains(KEGG) && !targetRefDB.toUpperCase().contains(ENSEMBL))
+		{
+			Optional<GKInstance> refDBInst = LinksToCheckCache.getCache().keySet().stream()
+																.filter( inst -> ((inst.getDisplayName().equals(targetRefDB) && LinksToCheckCache.getRefDBsToCheck().contains(targetRefDB))))
+																.findFirst();
```
cookersjs: Too much indenting here

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -514,20 +510,82 @@ else if (fileProcessorName.get() instanceof List<?>)
...
+			if (refDBInst.isPresent())
+			{
+				line = checkLinksForRefDB(new LinkCheckManager(), refDBInst.get());
+			}
+			else
+			{
+				line = targetRefDB + "\tN/A\tN/A\n";
+			}
+			return line;
```
cookersjs: What situation exists where the RefDB isn't present that is OK?  Wouldn't we want to report it rather than having a line of 'N/A's?

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -745,13 +803,15 @@ private String getUpToDateAccessURL(String resourceIdentifier, String accessURL)
...
+			final String reactomeIDToken = "###ID###";
+			if (!urlFromIdentifiersDotOrg.replace(identifiersDotOrgIDToken, "").equals(accessURL.replace(reactomeIDToken, "")))
 			{
 				// If replacing the Identifier tokens cause the two strings to mis-match, we should
 				// use the new accessURL from identifiers.org, and log a message so someone will
 				// know to update reference-databases.xml
-				updatedAccessURL = urlFromIdentifiersDotOrg.replace("{$id}", "###ID###");
-				logger.info("accessURL changed from: '{}' to: '{}', as per the data at identifiers.org", accessURL, urlFromIdentifiersDotOrg.replace("{$id}", "###ID###"));
+				updatedAccessURL = urlFromIdentifiersDotOrg.replace(identifiersDotOrgIDToken, reactomeIDToken);
+				logger.info("accessURL changed from: '{}' to: '{}', as per the data at identifiers.org", accessURL, updatedAccessURL);
```
cookersjs: I never realized I should update the reference-databases.xml file based on this message -- perhaps that should in the the info message?

src/main/java/org/reactome/addlinks/linkchecking/LinkChecker.java
```diff
@@ -69,9 +69,9 @@ public LinkCheckInfo checkLink() throws HttpHostConnectException, IOException, E
 											.setRelativeRedirectsAllowed(true)
 											.setConnectionRequestTimeout(1000 * (int)this.timeout.getSeconds()).build();
```
cookersjs: The indents here (and from line 65) are off

src/main/java/org/reactome/addlinks/linkchecking/LinkChecker.java
```diff
@@ -97,13 +97,14 @@ public LinkCheckInfo checkLink() throws HttpHostConnectException, IOException, E
 				// we will *not* handle socket timeouts (inactivity that occurs after the connection has been established).
 				// we will *not* handle connection manager timeouts (time waiting for connection manager or connection pool).
 				long endtime = System.currentTimeMillis();
-				e.printStackTrace();
+//				e.printStackTrace();
```
cookersjs: Should this be removed or uncommented?

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: This report is a summary of counts for each resource: how many OK, how many not OK.

More detailed info about, possiblty for each URL, should be in the LinkChecker log or LinkCheckManager log.

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: Lines up fine in Eclipse. Github's tabsize is a bit too large.

SolomonShorser-OICR: You can add "?ts=4" to set the tabsize to 4 (I think it defaults to 8 for some strange reason):

https://github.com/reactome/add-links/blob/d8cbcde034e6729a1a8609152708503d6682488f/src/main/java/org/reactome/addlinks/AddLinks.java?ts=4#L555-L557

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: No. Some ReferenceDatabases aren't in the list of ReferenceDatabases to check. See referenceDatabasesToLinkCheck in application-context.xml.

Usually when I don't want to check a reference database, I'll add it to the list, but _commented out_, with another comment explaining. Here's an example:

```xml
<!--
HGNC pages are now loaded using JavaScript so a simple GET request is not enough to see if the identifier is on the page.
<value>HGNC</value> -->
```

src/main/java/org/reactome/addlinks/linkchecking/LinkChecker.java
SolomonShorser-OICR: Yes, github's default tabsize is too large. Appending "?ts=4" to the URL can fix it.

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: _Just_ getting a 200 response code isn't terribly useful. There was one resource (won't name names but it rhymes with "leg" ;) ) that would give a 200 response with a message containing the identifier. And that message? It said the identifier could not be found.

I'd really prefer to define success a little more narrowly than just receiving a 200 response code.

We could revisit this in the future if it becomes an issue, but it's a bit beyond the scope of this PR.

cookersjs: Yea fair enough, I saw after I'd written this comment how you whittle down the success to 'the response contains the identifier' && its '200 || 300ish' -- I just wanted to see what you had to sat about it

src/main/java/org/reactome/addlinks/AddLinks.java
cookersjs: Ohh I see, I didn't realize that this is how not including (or commenting out) the resource manifests in the code so that it doesn't check something

SolomonShorser-OICR:
@jweiser Any thoughts on this PR?

jweiser:
> @jweiser Any thoughts on this PR?

Sorry, I've been distracted.  I'll finish posting any comments by Monday morning!

SolomonShorser-OICR:
OK

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -193,96 +197,88 @@ public void doAddLinks() throws Exception
...
 	 */
-	private String checkLinks()
-	{
-		StringBuilder linkCheckReportLines = new StringBuilder("RefDBName\tNumOK\tNumNotOK\n");
-		// Now, check the links that were created to ensure that they are all valid.
-		LinkCheckManager linkCheckManager = new LinkCheckManager();
-		linkCheckManager.setDbAdaptor(this.dbAdapter);
-		// Filter by references database name.
-		for (GKInstance refDBInst : LinksToCheckCache.getCache().keySet() )
+	private String checkLinksForRefDB(LinkCheckManager linkCheckManager, GKInstance refDBInst)
```
jweiser: It looks like the JavaDoc params are different than the parameters in the method signature.  Could they be synchronized?  

Adding a description for the method, a brief description for each param and an `@return` annotation would be great, too.

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -193,96 +197,88 @@ public void doAddLinks() throws Exception
...
+	{
+		 int numLinkOK = 0;
+		 int numLinkNotOK = 0;
+		logger.info("Link-checking for database: {}", refDBInst.getDisplayName());
+		StringBuilder reportLine = new StringBuilder();
+		if (!refDBInst.getDisplayName().toUpperCase().contains(ENSEMBL))
+		{
+			reportLine.append(refDBInst.getDisplayName()).append("\t");
+		}
```
jweiser: Why does the report have a tab appended if the reference db's display name doesn't contain "ENSEMBL"?

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -193,96 +197,88 @@ public void doAddLinks() throws Exception
...
+		else
+		{
+			try
+			{
+				Optional<String> longerName = ((List<String>)refDBInst.getAttributeValuesList(ReactomeJavaConstants.name)).stream().filter(name -> name.length() > ENSEMBL.length()).sorted().findFirst();
+				reportLine.append(longerName.orElse(ENSEMBL)).append("\t");
```
jweiser: The details of how the longest name is found are pretty involved.  Could this be abstracted to a named function so the flow of the method doesn't get interrupted trying to understand how the longest name is found?  Also, the logging on exception could be handled in that function as well if there is a problem.

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -193,96 +197,88 @@ public void doAddLinks() throws Exception
...
+				e.printStackTrace();
+			}
+			catch (Exception e)
+			{
+				logger.error("Error!", e);
+				e.printStackTrace();
+			}
+
+		}
+		Map<String, LinkCheckInfo> results = linkCheckManager.checkLinks(refDBInst, new ArrayList<>(LinksToCheckCache.removeRefDBFromCache(refDBInst)), this.proportionToLinkCheck, this.maxNumberLinksToCheck);
```
jweiser: Could the ArrayList be stored in a temp called something like `linksToCheck`?  It wasn't clear to me what `new ArrayList<>(LinksToCheckCache.removeRefDBFromCache(refDBInst)` was on the first reading.

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -193,96 +197,88 @@ public void doAddLinks() throws Exception
...
+		// "results" is a map of DB IDs mapped to link-checking results, for each identifier.
+		for (Entry<String, LinkCheckInfo> entry : results.entrySet())
```
jweiser: It looks like you're only using the values of the results map and not the keys?  If so, you could do:
`for (LinkCheckInfo result : results.values())`

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -297,18 +293,18 @@ private String checkLinks()
 	private void reportsAfterAddLinks(CrossReferenceReporter xrefReporter, DuplicateIdentifierReporter duplicateIdentifierReporter, Map<String, Map<String, Integer>> preAddLinksReport) throws SQLException, IOException, Exception
 	{
 		logger.info("Counts of references to external databases currently in the database ({}), AFTER running AddLinks", this.dbAdapter.getConnection().getCatalog());
-		//reporter.printReport();
 		Map<String, Map<String,Integer>> postAddLinksReport = xrefReporter.createReportMap();
-		logger.info("\n"+xrefReporter.printReport(postAddLinksReport));
+		String report = xrefReporter.printReport(postAddLinksReport);
```
jweiser: `printReport` kind of sounds like a method with a void return type that outputs to the screen.  Does the method do something in addition to creating the String that's returned?  If not, maybe `createReport` is a more precise name.

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -416,6 +412,8 @@ private void purgeUnusedRefDBs()
 	@SuppressWarnings("unchecked")
 	private void createReferences(long personID, Map<String, Map<String, ?>> dbMappings) throws IOException, Exception
 	{
+		String linkCheckReportName = LINK_CHECK_REPORTS_PATH + "/linkCheckSummaryReport" + DateTimeFormatter.ofPattern(DATE_PATTERN_FOR_FILENAMES).format(LocalDateTime.now()) + ".tsv";
+		Files.write(Paths.get(linkCheckReportName), "RefDBName\tNumOK\tNumNotOK\n".getBytes(), StandardOpenOption.CREATE);
```
jweiser: I know it's only two separators, but if more fields were to be added, it might be better and more readable to do:
`String.join("\t", "RefDBName", "NumOK", "NumNotOK")`

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -514,20 +512,82 @@ else if (fileProcessorName.get() instanceof List<?>)
 						}
 					}
 				}
-
+				if (refCreator instanceof NCBIGeneBasedReferenceCreator || refCreator instanceof ENSMappedIdentifiersReferenceCreator)
+				{
+					for (EntrezGeneBasedReferenceCreator subCreator : ((NCBIGeneBasedReferenceCreator)refCreator).getSubCreators())
```
jweiser: I'm a bit confused about why `ENSMappedIdentifiersReferenceCreator` is checked as a possible instance type if the refCreator is cast to `NCBIGeneBasedReferenceCreator` in the for loop?

src/main/java/org/reactome/addlinks/AddLinks.java
jweiser: Should a StandardOpenOption be added to handle the case if the file already exists (i.e. overwrite or append to the end of the file)?

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -514,20 +512,82 @@ else if (fileProcessorName.get() instanceof List<?>)
...
+	 */
+	private String checkLinksForRefCreator(BatchReferenceCreator<?> refCreator)
+	{
+		String targetRefDB = refCreator.getTargetRefDB();
+		String line = "";
+		if (!targetRefDB.toUpperCase().contains(KEGG) && !targetRefDB.toUpperCase().contains(ENSEMBL))
+		{
+			Optional<GKInstance> refDBInst = LinksToCheckCache.getCache().keySet().stream()
+																.filter( inst -> ((inst.getDisplayName().equals(targetRefDB) && LinksToCheckCache.getRefDBsToCheck().contains(targetRefDB))))
+																.findFirst();
```
jweiser: Could this be a separate method with a name describing how the refDBInst is obtained rather than showing the algorithm details?  Maybe something like `getTargetRefDbInstanceFromCache`?

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -514,20 +512,82 @@ else if (fileProcessorName.get() instanceof List<?>)
...
+		else if (targetRefDB.toUpperCase().contains(ENSEMBL))
+		{
+			StringBuilder sb = new StringBuilder();
+			Set<GKInstance> refDBInsts = LinksToCheckCache.getCache().keySet().stream().filter(inst -> inst.getDisplayName().toUpperCase().contains(ENSEMBL)).collect(Collectors.toSet());
+			for (GKInstance refDB : refDBInsts)
+			{
+				sb.append(checkLinksForRefDB(new LinkCheckManager(), refDB));
 			}
+			line = sb.toString();
 		}
```
jweiser: The logic in the `if` and `else if` blocks are identical except for the String value checked for in the `contains`.  Could the logic be moved to a method that accepts the name of the `targetRefDB` and returns the line of interest as a String?

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -634,53 +694,53 @@ private void executeCreateReferenceDatabases(long personID)
 	{
 		ReferenceDatabaseCreator creator = new ReferenceDatabaseCreator(this.dbAdapter, personID);
 
-		for (String key : this.referenceDatabasesToCreate.keySet())
+		for (Entry<String, Map<String, ?>> entry : this.referenceDatabasesToCreate.entrySet())
```
jweiser: Similar to above, it looks like the keys of the entry set are never used?  Could this be:
`for (Map<String, ?> refDB : this.referenceDatabasesToCreate.values())`

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -745,13 +805,15 @@ private String getUpToDateAccessURL(String resourceIdentifier, String accessURL)
...
+			final String reactomeIDToken = "###ID###";
+			if (!urlFromIdentifiersDotOrg.replace(identifiersDotOrgIDToken, "").equals(accessURL.replace(reactomeIDToken, "")))
 			{
 				// If replacing the Identifier tokens cause the two strings to mis-match, we should
 				// use the new accessURL from identifiers.org, and log a message so someone will
 				// know to update reference-databases.xml
-				updatedAccessURL = urlFromIdentifiersDotOrg.replace("{$id}", "###ID###");
-				logger.info("accessURL changed from: '{}' to: '{}', as per the data at identifiers.org", accessURL, urlFromIdentifiersDotOrg.replace("{$id}", "###ID###"));
+				updatedAccessURL = urlFromIdentifiersDotOrg.replace(identifiersDotOrgIDToken, reactomeIDToken);
+				logger.info("accessURL changed from: '{}' to: '{}', as per the data at identifiers.org - you may want to updated reference-databases.xml with the new accessURL.", accessURL, updatedAccessURL);
```
jweiser: Typo - "update"?

src/main/java/org/reactome/addlinks/linkchecking/LinkChecker.java
```diff
@@ -97,13 +97,13 @@ public LinkCheckInfo checkLink() throws HttpHostConnectException, IOException, E
...
 				// we will *not* handle connection manager timeouts (time waiting for connection manager or connection pool).
 				long endtime = System.currentTimeMillis();
-				e.printStackTrace();
 				this.numRetries++;
 				this.timeout = this.timeout.plus(Duration.ofSeconds(30));
 				logger.info("Failed to check {} after {} due to cause \"{}\", but will retry {} more time(s) with a longer timeout of {}.", this.uri, Duration.ofMillis(endtime - startTime), e.getMessage(), MAX_NUM_RETRIES - this.numRetries, this.timeout);
 				done = this.numRetries > MAX_NUM_RETRIES;
 				if (done)
 				{
+					e.printStackTrace();
```
jweiser: Should this be logged in addition to printing a stack trace?

src/main/java/org/reactome/addlinks/referencecreators/NCBIGeneBasedReferenceCreator.java
```diff
@@ -9,7 +11,14 @@
 public abstract class NCBIGeneBasedReferenceCreator extends SimpleReferenceCreator< Map<String,List<String>> >
 {
 	protected List<EntrezGeneBasedReferenceCreator> entrezGeneReferenceCreators;
-	
+
+	public List<EntrezGeneBasedReferenceCreator> getSubCreators()
+	{
+		return this.entrezGeneReferenceCreators != null
+				? Collections.unmodifiableList(this.entrezGeneReferenceCreators)
+				: Collections.unmodifiableList(new ArrayList<EntrezGeneBasedReferenceCreator>());
```
jweiser: For the last part of the ternary, you could say `Collections.emptyList()`.  I think the compiler should be able to infer the type and `emptyList()` will return an immutable list: https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#emptyList--

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: ENSEMBL names require special handling.

IF the ReferenceDatabase is _NOT_ ENSEMBL, append the reference database name and a tab to the current report line. The tab is because the output is a TSV file, so after the value of the report name, there must be a tab.
ELSE (the ReferenceDatabase _is_ ENSEMBL), find a longer name because ENSEMBL ReferencesDatabases have multiple names. All of them have the name "ENSEMBL", but for the report, we want a name that's species-specific. The easiest way to do that is to find a name that is longer than the the string "ENSEMBL". Append _that_ name to the report line, with a tab.

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: The file name should have the datetime, specific down to the second, it is unlikely there will be another file already there with the same name.

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: I could have cast to either `ENSMappedIdentifiersReferenceCreator` or `ENSMappedIdentifiersReferenceCreator`. Either one should work, but I have to choose one.

SolomonShorser-OICR: ...but if I _don't_ cast, then I don't get access to `getSubCreators`.

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -694,14 +699,13 @@ private void executeCreateReferenceDatabases(long personID)
 	{
 		ReferenceDatabaseCreator creator = new ReferenceDatabaseCreator(this.dbAdapter, personID);
 
-		for (Entry<String, Map<String, ?>> entry : this.referenceDatabasesToCreate.entrySet())
+		for (Map<String, ?> refDBMapping : this.referenceDatabasesToCreate.values())
```
cookersjs: It's not a new thing, but maybe a comment on what the ? object type could be? Or a general overview of what the object could be?

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: Not really a fan of that refactoring,

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: If it gets longer, I'll probably do that.

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: Not a fan of this refactoring.

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -143,7 +144,7 @@ public void doAddLinks() throws Exception
 		}
 		// Start by creating ReferenceDatabase objects that we might need later.
 		this.executeCreateReferenceDatabases(personID);
-		// Now that we've *created* new ref dbs, rebuild any caches that might have depended on them.
+		// Now that we've *created* new ref dbs, rebuild any caches that might have dependended on them.
```
cookersjs, line 147: depended typo

src/main/java/org/reactome/addlinks/fileprocessors/HmdbMetabolitesFileProcessor.java
```diff
@@ -132,7 +130,7 @@ public HmdbMetabolitesFileProcessor()
 	
 	private void transformXmlToTsv(String inputFilename, String outfileName)
 	{
-		
+		System.setProperty("javax.xml.transform.TransformerFactory", "com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl");
```
cookersjs, line 133: Set the two strings here as final variables?

SolomonShorser-OICR: Unlikely they will ever be used anywhere else.

cookersjs: Fair point

pom.xml
```diff
@@ -225,6 +225,9 @@
 						<manifest>
 							<mainClass>org.reactome.addlinks.Main</mainClass>
 						</manifest>
+						<manifestEntries>
+							<Multi-Release>true</Multi-Release>
+						</manifestEntries>
```
jweiser, line 230: I'm not familiar with this tag - what does it do or convey?

src/main/java/org/reactome/addlinks/AddLinks.java
jweiser: My question here is more why is `ENSMappedIdentifiersReferenceCreator` checked as a possibility if the cast on line 122 is going to be `NCBIGeneBasedReferenceCreator`?  Would the first condition on line 520 `refCreator instanceof NCBIGeneBasedReferenceCreator` be enough and the `||` not needed?

pom.xml
SolomonShorser-OICR: For Java versions > 8, this specifies a mulit-release jar:

https://in.relation.to/2017/02/13/building-multi-release-jars-with-maven/

https://stackoverflow.com/questions/60741008/maven-assembly-plugin-multi-releasetrue

I guess I needed it to get something running properly here. I can remove it, though if Justin's already run it on the Release server, it's unlikely to cause problems.

src/main/java/org/reactome/addlinks/AddLinks.java
SolomonShorser-OICR: There is no cast on line 122... :/

SolomonShorser-OICR: Because `refCreator` could be either an `NCBIGeneBasedReferenceCreator`, _or_ it could be an `ENSMappedIdentifiersReferenceCreator`. When it comes time to use it, it doesn't matter _which_ we cast to - as long as we cast to _something_ that will give access to the `getSubCreators` method.

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -513,20 +517,88 @@ else if (fileProcessorName.get() instanceof List<?>)
...
+																.findFirst();
+			if (refDBInst.isPresent())
+			{
+				line = checkLinksForRefDB(new LinkCheckManager(), refDBInst.get());
 			}
+			else
+			{
+				line = targetRefDB + "\tN/A\tN/A\n";
+			}
+			return line;
```
jweiser, line 573: Could this be reduced to
```
String line = refDBInst.isPresent() ? 
    checkLinksForRefDB(new LinkCheckManager(), refDBInst.get()) : 
    targetRefDB + "\tN/A\tN/A\n"`;
```

SolomonShorser-OICR: Probably, but I don't like ternaries that much. I find this suggested replacement a little hard to read. I am usually OK with ternaries when the expressions are much shorter.

SolomonShorser-OICR: I mean... it could be reduced even further to 
```Java
return refDBInst.isPresent() ? 
          checkLinksForRefDB(new LinkCheckManager(), refDBInst.get()) : 
          targetRefDB + "\tN/A\tN/A\n"
```

But I find this _waaaaay_ too terse.

jweiser: Okay

pom.xml
jweiser: Okay, just wondering

src/main/java/org/reactome/addlinks/AddLinks.java
jweiser: Yeah, line 522 - sorry.  
That's true, but isn't `ENSMappedIdentifiersReferenceCreator` a sub-class of `NCBIGeneBasedReferenceCreator`?  So you only need to have the more general check of `refCreator instanceof NCBIGeneBasedReferenceCreator` on line 520.  If that's false, then the second check will never be true.

jweiser: That is my issue isn't with the cast - it's with the second `instanceof` check.

SolomonShorser-OICR: Hmm.... I believe this was was necessary in the past but I think the type hierarchy has changed so it may no longer be necessary. I could remove it, but that would require testing. Might not finish by the end of the day...

jweiser: Okay, it's not a blocker but just something I noticed that could be simplified.  I don't want to hold up approving and merging the PR for it, so maybe just a TODO comment to test it?

SolomonShorser-OICR: Testing is already underway.

src/main/java/org/reactome/addlinks/AddLinks.java
jweiser: ^-- What are your thoughts on this one?

SolomonShorser-OICR: Ok, I fixed this.

src/main/java/org/reactome/addlinks/AddLinks.java
```diff
@@ -242,7 +242,8 @@ private String checkLinksForRefDB(LinkCheckManager linkCheckManager, GKInstance
 			}
 
 		}
-		Map<String, LinkCheckInfo> checkedLinks = linkCheckManager.checkLinks(refDBInst, new ArrayList<>(LinksToCheckCache.removeRefDBFromCache(refDBInst)), this.proportionToLinkCheck, this.maxNumberLinksToCheck);
+		List<GKInstance> instancesToCheck = new ArrayList<>(LinksToCheckCache.removeRefDBFromCache(refDBInst));
```
cookersjs, line 245: Just curious about what this line in particular its doing -- its both removing refDBs from cache and generating a list of instances to check?

SolomonShorser-OICR: It calls the `remove` method on the underlying map. The `remove` method is like `get`, but the data is removed from the map.

PR #137, opened 2020-04-03
Memory fixes

SolomonShorser-OICR:
Fixes some memory-related issues:
 - Transform HMDB Metabolites XML file in chunks, rather than attempt to transform the whole file at once.
 - Fixed Ensembl Biomart file processor to process files line-by-line rather than trying to use `readAllLines`
 - some other tweaks and fixes.

### DIFF
--- a/src/main/java/org/reactome/addlinks/fileprocessors/HmdbMetabolitesFileProcessor.java
+++ b/src/main/java/org/reactome/addlinks/fileprocessors/HmdbMetabolitesFileProcessor.java
@@ -1,7 +1,12 @@
 package org.reactome.addlinks.fileprocessors;
 
 import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.FileReader;
+import java.io.FileWriter;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.Arrays;
@@ -10,17 +15,25 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLStreamConstants;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
 import javax.xml.transform.Result;
 import javax.xml.transform.Source;
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerConfigurationException;
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.stax.StAXSource;
 import javax.xml.transform.stream.StreamResult;
 import javax.xml.transform.stream.StreamSource;
 
 public class HmdbMetabolitesFileProcessor extends FileProcessor<Map<HmdbMetabolitesFileProcessor.HMDBFileMappingKeys, ? extends Collection<String>>>
 {
+	private static final Object METABOLITE_ELEMENT_NAME = "metabolite";
+
+
 	public HmdbMetabolitesFileProcessor(String processorName)
 	{
 		super(processorName);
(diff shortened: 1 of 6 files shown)


src/main/java/org/reactome/addlinks/fileprocessors/HmdbMetabolitesFileProcessor.java
```diff
@@ -130,4 +129,54 @@ public HmdbMetabolitesFileProcessor()
...
+			XMLStreamReader xsr = xif.createXMLStreamReader(new FileReader(inputFilename));
+			xsr.nextTag();
+			Source source = new StreamSource(this.getClass().getClassLoader().getResourceAsStream("hmdb_metabolites_transform.xsl"));
+			TransformerFactory factory = TransformerFactory.newInstance();
+			Transformer transformer = factory.newTransformer(source);
+			
+			while (xsr.nextTag() == XMLStreamConstants.START_ELEMENT)
+			{
+				if (xsr.getName().getLocalPart().equals(METABOLITE_ELEMENT_NAME))
+				{
```
cookersjs, line 148: It was pretty intuitive to me what was going on in this method until this part. Could you include a small description of what specifically happens once the 'metabolite' tag is found?

src/main/java/org/reactome/addlinks/fileprocessors/HmdbMetabolitesFileProcessor.java
```diff
@@ -130,4 +129,54 @@ public HmdbMetabolitesFileProcessor()
...
+		catch (TransformerException e)
+		{
+			logger.error("Data transformation error!", e);
+			e.printStackTrace();
+		}
+		catch (IOException e)
+		{
+			logger.error("I/O Error", e);
+			e.printStackTrace();
+		}
```
cookersjs, line 180: Should a final general Exception try/catch be put at the end?

SolomonShorser-OICR: Why? None of the methods being called throw `Exception`, they all throw more specific types of `Exception`.

jweiser, line 180: This is a lot of code in the method body for exception handling.  Since the only thing that differs is the logged error message, could you do this:

```
try {
    ...
} catch (FileNotFoundException | XMLStreamException | TransformerException | IOException e) {
    // Or use a helper method call to get the String you want
    logger.error("Error due to " + e.getClass().getName() + " exception", e);
    e.printStackTrace();
}
```
You can just have a generic message since the exception type will be included in the log or you can have a method like 'private String getExceptionMessage(Exception e)` that returns the String error message you want to log without distracting from the main logic.

SolomonShorser-OICR: I like having more specific messages. Gives the reader of the message a bit more of a clue than a generic message.

jweiser: That's fine - you can use a helper method to return the specific message you want.  I just think over 20 lines to handle exceptions is a bit much.

src/main/java/org/reactome/addlinks/fileprocessors/HmdbMetabolitesFileProcessor.java
SolomonShorser-OICR: Comments were added.

src/main/java/org/reactome/addlinks/fileprocessors/HmdbMetabolitesFileProcessor.java
SolomonShorser-OICR: I've removed the `e.printStackTrace()` calls (I don't think they were necessary, since the overload of the `error()` method should _log_ the full stack trace), but I'm still not in favour of a helper method here.

jweiser: Is there a reason you're opposed to a helper method?  

I think it would be useful - it allows you to collapse the exception handling into one catch block and reduce the size of the method.  I think it makes sense to try to keep the method concise and as focused on its main purpose as possible.  Using a helper method would help reduce the footprint of the exception handling which is incidental to the method's purpose.  For me, it really helps readability and understanding.

SolomonShorser-OICR: In this particular case, `FileNotFoundException` is a subtype of `IOException` so the expression `catch (FileNotFoundException | XMLStreamException | TransformerException | IOException e)` causes problems because `FileNotFoundException` _is_ an `IOException`. If you want to give them separate handling (and I did because they are not thrown by the same message so I wanted different messges), they need separate `catch`es. So I'd have to have at least two separate catches - one for `FileNotFoundException`, the other for the remaining three that are handled by a helper method, which I personally would find confusing and weird. Or, lose the distinction between `FileNotFound...` and `IO...`.

AND... there's a similar problem with `TransformerConfigurationException` and `TransformerException`... I prefer to keep them separate so I can give a specific messages because one exception is caused by the transformation itself, and the other is caused by issues within the XSL file.

So... in order to keep the 5 distinct messages, I still need 3 `catch`-blocks:
```Java
catch (FileNotFoundException e)
{
    logger.error("Could not read input file", e);
}
catch (XMLStreamException| TransformerConfigurationException | IOException e) 
{
    logXml2TsvException(e);
}
catch (TransformerException e)
{
    logger.error("Data transformation error", e);
}
```
And the handler will only handle `XMLStreamException`, `TransformerConfigurationException`, `IOException`.

---
In general, I think that 4 or 5 or 6 `catch`-blocks with _simple_ exception-handling code is really not at all unreasonable. In this case, I think adding another method to handle all exceptions will increase overall complexity - now the method needs to have its own logic to determine which exception type it received so it knows what to do whereas the original try/catch had that logic built in.

jweiser: Okay, I see what you're saying.  We can leave it.

PR #139, opened 2020-06-01
Feature/73 changes

cookersjs:
- NCBI IDs are added to OtherIdentifiers via AddLinks (NCBI already has Xreferences on every Reactome details page, but formatted as `NCBI Gene: ${identifier} ${gene}`, but the only problem with analysis core is it has no way of looking for just the identifier, because the gene is in every cross reference displayName. Rather than create a redundant cross reference, it was much easier to just add one more value to the otherIdentifiers slot)
- Canis familiaris Ensembl database no longer valid -- needs to include 'lupus'
- BioMart downlaod for `cfamiliaris` changed to be `clfamiliaris`
- Fix to how biomart files are saved
- s3 integration for Addlinks downloads and insertions
- CTD URL updated
- FlyBase URL updated


src/main/java/org/reactome/addlinks/EnsemblBioMartUtil.java
```diff
@@ -66,7 +65,13 @@ private EnsemblBioMartUtil() {}
...
      */
     public static String getBioMartSpeciesName(String speciesName) {
-        return speciesName.substring(0,1).toLowerCase() + speciesName.split(" ")[1];
+        String biomartSpeciesName = speciesName.substring(0,1).toLowerCase() + speciesName.split(" ")[1];
+        // May 2020 -- BioMart changed the Dog BioMart name from 'cfamiliaris' to 'clfamiliaris'.
+        // This can't be done automatically, so needs to be adjusted specifically.
+        if (speciesName.equals("Canis familiaris")) {
+            biomartSpeciesName = speciesName.substring(0,1).toLowerCase() + "l" + speciesName.split(" ")[1];
+        }
+        return biomartSpeciesName;
```
jweiser, line 74: Maybe for clarity in the speciesName processing, you could add these two temp variables to be used to form the `biomartSpeciesName` (renaming the parameter to something like `fullSpeciesName` or `genusAndSpeciesName`:

```
if (genusAndSpeciesName.equals("Canis familiaris")) {
   return "clfamiliaris"; // You could construct the name programatically, but maybe a literal String is better so it sticks out that this is an anamoly and an exception to the pattern
}

String genusAbbreviation = genusAndSpeciesName.substring(0,1).toLowerCase() + speciesName.split(" ")[1];
String speciesName = genusAndSpeciesName.split(" ");

return genusAbbreviation + speciesName;
```

src/main/java/org/reactome/addlinks/dataretrieval/ensembl/EnsemblBioMartRetriever.java
```diff
@@ -61,27 +63,32 @@ public void downloadData() throws IOException, InterruptedException, BioMartQuer
         // Species names are in bioMart format (eg: hsapiens).
         for (String speciesName : EnsemblBioMartUtil.getSpeciesNames()) {
             String speciesBioMartName = EnsemblBioMartUtil.getBioMartSpeciesName(speciesName);
-            logger.info("Retrieving BiomMart files for " + speciesBioMartName);
-
+            logger.info("Retrieving BioMart files for " + speciesBioMartName);
+            Set<String> biomartSearchTerms = new HashSet<>(Arrays.asList(GO_ID_BIOMART_SEARCH_TERM, GO_SLIM_BIOMART_SEARCH_TERM, NCBI_ENTREZ_BIOMART_SEARCH_TERM));
```
jweiser, line 67: I'm looking forward to when we are using Java 9+ and can just do `Set.of()` :).

src/main/java/org/reactome/addlinks/dataretrieval/ensembl/EnsemblBioMartRetriever.java
```diff
@@ -61,27 +63,32 @@ public void downloadData() throws IOException, InterruptedException, BioMartQuer
...
-            for (String microarrayType : microarrayTypes) {
-                queryBioMartAndStoreData(speciesBioMartName, getBioMartXMLFilePath(), microarrayType, EnsemblBioMartUtil.MICROARRAY_IDS_AND_GO_TERMS_SUFFIX);
-            }
-
-            // Query GO and GO_SLIM identifier mapping data from BioMart and write it to a file (eg: hsapiens_microarray_ids_and_go_terms)
-            logger.info("Retrieving GO data");
-            for (String goQueryId : Arrays.asList(GO_ID_BIOMART_SEARCH_TERM, GO_SLIM_BIOMART_SEARCH_TERM)) {
-                queryBioMartAndStoreData(speciesBioMartName, getBioMartXMLFilePath(), goQueryId, EnsemblBioMartUtil.MICROARRAY_IDS_AND_GO_TERMS_SUFFIX);
+            biomartSearchTerms.addAll(queryBioMart(getMicroarrayTypesQuery(speciesBioMartName), MICROARRAY_TYPES));
+            // Iterate through each Biomart search term and retrieve Ensembl-otherIdentifier identifier mappings.
```
jweiser, line 71: Nitpick :) - BioMart

src/main/java/org/reactome/addlinks/dataretrieval/ensembl/EnsemblBioMartRetriever.java
```diff
@@ -61,27 +63,32 @@ public void downloadData() throws IOException, InterruptedException, BioMartQuer
...
-            for (String uniprotQueryId : Arrays.asList(UNIPROT_SWISSPROT_BIOMART_SEARCH_TERM, UNIPROT_TREMBL_BIOMART_SEARCH_TERM)) {
-                queryBioMartAndStoreData(speciesBioMartName, getBioMartXMLFilePath(), uniprotQueryId, EnsemblBioMartUtil.UNIPROT_SUFFIX);
+            String uniprotFilename = this.destination + speciesBioMartName + EnsemblBioMartUtil.UNIPROT_SUFFIX;
+            if (!Files.exists(Paths.get(uniprotFilename))) {
+                logger.info("Retrieving UniProt mappings from BioMart");
+                for (String uniprotQueryId : Arrays.asList(UNIPROT_SWISSPROT_BIOMART_SEARCH_TERM, UNIPROT_TREMBL_BIOMART_SEARCH_TERM)) {
+                    queryBioMartAndStoreData(speciesBioMartName, getBioMartXMLFilePath(), uniprotQueryId, uniprotFilename);
+                }
+            } else {
+                logFileExistsMessage(uniprotFilename);
```
jweiser, line 91: The logic and structure of this block of code are almost identical to the BioMart search terms on lines 73-80.  Could you consolidate it in a more generic helper method that takes in the differences of the appropriate suffix (lines 73 and 84) and list of search terms (lines 75 and 87) as parameters?

src/main/java/org/reactome/addlinks/dataretrieval/ensembl/EnsemblBioMartRetriever.java
```diff
@@ -98,26 +105,20 @@ private String getMicroarrayTypesQuery(String bioMartSpeciesName) {
...
-                e.printStackTrace();
-            }
-            storeBioMartData(biomartFilename, biomartResponseLines);
-        }
-        else {
-        	logger.info("{} already exists. It is *assumed* to be complete, and will not be downloaded.", biomartFilename);
+        Set<String> biomartResponseLines = new HashSet<>();
+        logger.info("Retrieving data associated with query ID: {}", biomartDataType);
+        try {
+            biomartResponseLines = queryBioMart(getBioMartIdentifierQuery(biomartQueryFilePath, biomartSpeciesName, biomartDataType), biomartDataType);
```
jweiser, line 116: Since this is a method call using the result of another method call, could this be broken up on a few lines and indented so it's clearer at a glance which method call is getting which parameters?

src/main/java/org/reactome/addlinks/dataretrieval/ensembl/EnsemblBioMartRetriever.java
cookersjs: Yep, good catch

src/main/java/org/reactome/addlinks/EnsemblBioMartUtil.java
cookersjs: Sure, I'll do that

src/main/java/org/reactome/addlinks/dataretrieval/ensembl/EnsemblBioMartRetriever.java
```diff
@@ -259,6 +260,11 @@ private String joinQueryFileLines(String pathToFile) throws IOException {
         return String.join(System.lineSeparator(), fileLines);
     }
 
+    // Message output whenever a file already exists.
```
SolomonShorser-OICR, line 263: Please try to keep method-level documentation written as Javadoc.

PR #146, opened 2020-11-17
Feature/update cosmic url

SolomonShorser-OICR:
Download a newer COSMIC file, and some code changes to support it:

 - file is too big to be buffered in memory so write to output stream when downloading
 - specify character set for file

Minor, unrelated change:
 - upgrade JUnit dependency from 4.12 to 4.13.1 b/c of dependabot's nagging. Ran some JUnit tests, they still run fine.

### DIFF
--- a/src/main/java/org/reactome/addlinks/fileprocessors/COSMICFileProcessor.java
+++ b/src/main/java/org/reactome/addlinks/fileprocessors/COSMICFileProcessor.java
@@ -1,11 +1,13 @@
 package org.reactome.addlinks.fileprocessors;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
 
 public class COSMICFileProcessor extends FileProcessor<String>
 {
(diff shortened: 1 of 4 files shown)


src/main/java/org/reactome/addlinks/fileprocessors/COSMICFileProcessor.java
```diff
@@ -23,17 +25,23 @@ public COSMICFileProcessor(String processorName)
...
+			try(Stream<String> lineStream = Files.lines(Paths.get(unzippedFile), StandardCharsets.ISO_8859_1))
 			{
-				//I added ".sequential()" because I have line counter and I want to ensure that it works properly.
-				Files.lines(Paths.get(unzippedFile)).filter(p -> !p.startsWith("#")).sequential().forEach( line ->
+				// Use parallel streams at your own peril! I kept running out of RAM and
+				// crashing on my OICR laptop - the server or the office-based workstation can
+				// *probably* handle it but since I can't successfully test, I will leave this
+				// as a sequential stream, for now. But honestly, it took my laptop ~7 minutes
+				// to process the file. I can't imagine a parallel stream would save more
+				// than 3 or 4 minutes...
```
cookersjs, line 41: Is this newer COSMIC file much bigger? I don't notice the file processing for the current cosmic file taking anywhere near as long as this

src/main/resources/basic-file-retrievers.xml
```diff
@@ -166,7 +166,7 @@
 			<bean class="org.reactome.addlinks.dataretrieval.COSMICFileRetriever" id="COSMIC">
 				<property name="userName" value="${cosmic.user}"/>
 				<property name="password" value="${cosmic.password}"/>
-				<property name="dataURL" value="https://cancer.sanger.ac.uk/cosmic/file_download/GRCh38/cosmic/v87/CosmicMutantExport.tsv.gz" />
+				<property name="dataURL" value="https://cancer.sanger.ac.uk/cosmic/file_download/GRCh38/cosmic/v92/CosmicMutantExport.tsv.gz" />
```
cookersjs, line 169: This is a similar problem as FlyBase: regular file name changes that aren't mapped to something generic like 'current'. For FlyBase, my idea for Jenkins integration was for a user input question asking for the date string in the URL and then using a grep substitution with the value in the config file (this idea can be seen in use at the new orthoinference JF: [here](https://github.com/reactome/release-orthoinference/blob/feature/74_changes/Jenkinsfile#L58), which preprends the species ID on the log files). 

Something similar could be done for COSMIC in the future. Maybe we need to start listing variable resource names.

cookersjs (approved):

src/main/java/org/reactome/addlinks/fileprocessors/COSMICFileProcessor.java
SolomonShorser-OICR: I can't recall how previous one was, but yes. This newer one is big.

src/main/resources/basic-file-retrievers.xml
SolomonShorser-OICR: It would also need to work _outside_ of a Jenkins context.

cookersjs: Well, at the moment we just manually update FlyBase URLs, so I was more of thinking that automatic update functionality would just be bonus functionality for Jenkins

src/main/java/org/reactome/addlinks/fileprocessors/COSMICFileProcessor.java
```diff
@@ -23,17 +25,23 @@ public COSMICFileProcessor(String processorName)
...
 	{
-		Map<String,String> mappings = new HashMap<String, String>();
+		Map<String,String> mappings = new ConcurrentHashMap<>();
 		try
 		{
 			String unzippedFile = this.unzipFile(this.pathToFile);
 			AtomicInteger lineCount = new AtomicInteger(0);
-			try
+
+			try(Stream<String> lineStream = Files.lines(Paths.get(unzippedFile), StandardCharsets.ISO_8859_1))
```
jweiser, line 34: Just wondering, why the ISO_8859_1Character Set instead of UTF_8?

jweiser (approved):

src/main/java/org/reactome/addlinks/fileprocessors/COSMICFileProcessor.java
SolomonShorser-OICR: Because it worked.

PR #147, opened 2020-11-17
Feature/additional-logging

cookersjs:
An incredibly small PR that just adds some informational logging messages. Due to time constraints on my end, there are a couple bigger tasks that remain to be done between v75 and v76.

### DIFF
--- a/src/main/resources/addlinks.properties
+++ b/src/main/resources/addlinks.properties
@@ -1,5 +1,5 @@
 filterFileRetrievers=true
-executeAsPersonID=8939149
+executeAsPersonID=
 numberOfUniprotDownloadThreads=20
 lazyLoadCache=true
 proportionToLinkCheck=0.1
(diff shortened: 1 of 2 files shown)


SolomonShorser-OICR (approved):

src/main/resources/addlinks.properties
```diff
@@ -1,9 +1,9 @@
 filterFileRetrievers=true
-executeAsPersonID=8939149
+executeAsPersonID=
```
jweiser, line 2: Just curious why the person id was removed?

jweiser (approved):

src/main/resources/addlinks.properties
cookersjs: Just because it was hard-coded. All of our past AddLinks releases have been 'done' by Solomon as a result instead of the person running it

PR #155, opened 2021-01-26
Feature/pharmaco db

### LINKED ISSUE #154: Create references to PharmacoDB
Create links to http://pharmacodb.ca

Use provided files to link Reactome entities to PharmacoDB entities. Might need to use IUPHAR file (https://www.guidetopharmacology.org/DATA/ligands.csv) to map PubChem identifiers, which will be mapped to PharmacoDB identifiers (they will provide the file, at least until web API is online).

SolomonShorser-OICR:
This PR resolves #154 

Highlights:
 - Create references to PharmacoDB
 - Create caches of ReferenceTherapeutics in ReferenceObjectCache, to support creation of references to PharamcoDB.


src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
```diff
@@ -42,13 +42,13 @@ public ReferenceObjectCache(MySQLAdaptor adapter, boolean lazyLoad)
 	{
 		ReferenceObjectCache.lazyLoad = lazyLoad;
```
cookersjs, line 43: Is this `lazyLoad` variable still being used anywhere?

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
```diff
@@ -106,183 +106,217 @@ private static synchronized void buildReferenceCaches(String className, Map<Stri
 				logger.error("Could not get the \"{}\" attribute for {}. Reason: {}", identifierAttribute, referenceObject, e.getMessage());
 			}
 
-			// ReferenceMolecules and DatabaseIdentifiers do not have associated species.
-			if ( !className.equals(ReactomeJavaConstants.ReferenceMolecule) && !className.equals(ReactomeJavaConstants.DatabaseIdentifier) )
+			// ReferenceMolecules and DatabaseIdentifiers and ReferenceTherapeutics do not have associated species.
+			if ( !className.equals(ReactomeJavaConstants.ReferenceMolecule) && !className.equals(ReactomeJavaConstants.DatabaseIdentifier) && !className.equals(ReactomeJavaConstants.ReferenceTherapeutic))
```
cookersjs, line 110: If this block is only trying to filter out classes where species isn't valid, you can also use `instance.getSchemClass().isValidAttribute(ReactomeJavaConstants.species)`

cookersjs: If there is a `GKInstance` available, I should say

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
```diff
@@ -106,183 +106,217 @@ private static synchronized void buildReferenceCaches(String className, Map<Stri
...
 			if (cacheByRefDB != null)
 			{
-				String refDBID;
+				addObjectToRefDBCache(cacheByRefDB, referenceObject);
+			}
+			// DB_ID Cache
+			cacheByID.put(String.valueOf(referenceObject.getDBID()), referenceObject);
+			
+			// Identifier cache
+			if ( !className.equals(ReactomeJavaConstants.Reaction) )
```
cookersjs, line 124: Why is this for `Reaction` and not `ReactionlikeEvent`?

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
```diff
@@ -106,183 +106,217 @@ private static synchronized void buildReferenceCaches(String className, Map<Stri
...
+			catch (Exception e)
+			{
+				logger.error("Could not get the \"{}\" attribute for {}. Reason: {}", ReactomeJavaConstants.species, referenceObject, e.getMessage());
 			}
-			// DB_ID Cache
-			cacheByID.put(String.valueOf(referenceObject.getDBID()), referenceObject);
 			
-			// Identifier cache
-			if ( !className.equals(ReactomeJavaConstants.Reaction) )
+			if (null == species)
```
cookersjs, line 250: Perhaps a comment for what will specifically happen in cases where the `species` object is null? To me it looks like it iterates through `referenceTranscript` referrals to try and get any species, but some explanation would help

cookersjs (approved):
Looks good! Seems like the changes to the caching were a lot more involved then the actual PharmacoDB xref creation part

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
SolomonShorser-OICR: I liked listing them explicitly, to make it more clear, but if I have to add any more, I might switch to the `isValidAttribute` method.

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
SolomonShorser-OICR: We only create references on Reactions for Rhea and TargetPathogen, and I guess when it was discussed, it was decided to use `Reaction`.

cookersjs: Also I just realized that if you are in fact trying to determine broadly if its a `ReactionlikeEvent`, rather than specifically a `Reaction`, the `instance.getSchemClass().isa(ReactomeJavaConstants.ReactionlikeEvent`  approach will be needed to get proper coverage

SolomonShorser-OICR:
> Looks good! Seems like the changes to the caching were a lot more involved then the actual PharmacoDB xref creation part

Actually, most of the changes to the caching system were mostly refactoring and cleanup, not major logical changes. ReferenceTherapeutic objects are now included in the cache process, but that itself was not much work.

SolomonShorser-OICR:
@jweiser Any thoughts? I think this one is pretty much complete, as PharmacoDB has sent the URL for their data file and their new site is out of beta, so we should be able to run this for the current Release 76.

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
```diff
@@ -106,183 +106,217 @@ private static synchronized void buildReferenceCaches(String className, Map<Stri
...
 				try
 				{
-					refDBID = String.valueOf(((GKInstance) referenceObject.getAttributeValue(ReactomeJavaConstants.referenceDatabase)).getDBID());
-					if (! cacheByRefDB.containsKey(refDBID))
+					if (!objectCacheByIdentifier.containsKey(identifier))
 					{
-						List<GKInstance> byRefDb = Collections.synchronizedList( new LinkedList<GKInstance>() );
-						byRefDb.add(referenceObject);
-						cacheByRefDB.put(refDBID, byRefDb);
+						objectCacheByIdentifier.put(identifier, Collections.synchronizedList(new ArrayList<GKInstance>(Arrays.asList(referenceObject))) );
```
jweiser, line 162: I think `Arrays.asList` will return an ArrayList, so it should be safe to remove the wrapping of `new ArrayList<GKInstance>`

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
```diff
@@ -106,183 +106,217 @@ private static synchronized void buildReferenceCaches(String className, Map<Stri
...
+		}
+		catch (InvalidAttributeException e)
+		{
+			logger.error("Object {} does not have an identifier attribute, error: {}", referenceObject, e.getMessage());
+			e.printStackTrace();
+		}
+		catch (Exception e)
+		{
+			e.printStackTrace();
+			throw new Error(e);
```
jweiser, line 186: Is there a need to print the stacktrace if the error is rethrown (i.e. once the error is handled by Java, a stacktrace should be printed)?

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
```diff
@@ -106,183 +106,217 @@ private static synchronized void buildReferenceCaches(String className, Map<Stri
...
+	 * @param referenceObject - the object to add.
+	 */
+	private static void addObjectToSpeciesCache(String className, Map<String, List<GKInstance>> cacheBySpecies, GKInstance referenceObject)
+	{
+		// Sets do not allow duplicates.
+		Set<String> allSpecies = null;
+		//ReferenceRNASequence objects don't always have Species info, for some reason, so we need to get that from the associated ReferenceGeneProduct
+		if (className.equals(ReactomeJavaConstants.ReferenceRNASequence))
+		{
+			Object species = null;
```
jweiser, line 236: Is there a reason this is an `Object` instead of a `GKInstance`?

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
```diff
@@ -106,183 +106,217 @@ private static synchronized void buildReferenceCaches(String className, Map<Stri
...
+	 * Adds an object to a cache that is indexed by species.
+	 * @param className - the className of the object.
+	 * @param cacheBySpecies - the cache to add the object to.
+	 * @param referenceObject - the object to add.
+	 */
+	private static void addObjectToSpeciesCache(String className, Map<String, List<GKInstance>> cacheBySpecies, GKInstance referenceObject)
+	{
+		// Sets do not allow duplicates.
+		Set<String> allSpecies = null;
+		//ReferenceRNASequence objects don't always have Species info, for some reason, so we need to get that from the associated ReferenceGeneProduct
```
jweiser, line 233: This might be worth bringing up to the curators on a Tuesday call to make sure it's not a data error.  If all ReferenceRNASequences should have species, there should probably be a QA check.

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
```diff
@@ -106,183 +106,217 @@ private static synchronized void buildReferenceCaches(String className, Map<Stri
...
+			{
+				logger.error("Could not get the \"{}\" attribute for {} because it not valid for this object.", ReactomeJavaConstants.species, referenceObject);
+			}
+			catch (Exception e)
+			{
+				logger.error("Could not get the \"{}\" attribute for {}. Reason: {}", ReactomeJavaConstants.species, referenceObject, e.getMessage());
+			}
+			
+		}
+		if (cacheBySpecies != null)
```
jweiser, line 303: Would this mean the method would not cache the object if the parameter is null?

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
SolomonShorser-OICR: The code at this point doesn't use the _value_ of the `species` variable, it only is interested in whether or not a referenceObject _has_ a value for its `species` attribute. Since `getAttributeValue` returns `Object`, and since I don't actually _need_ to know anything more than is it null/non-null, I used `Object`. If the type of `species` is `GKInstance`, then I need to case the return of `getAttributeValue` to `GKInstance`, which seems silly since I'm not even going to use it as a `GKInstance`.

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
SolomonShorser-OICR: The object will not get added to a species-based cache. This is done with ReferenceMolecules and ReferenceTherapeutics and DatabaseIdentifiers, because they do not have a species attribute, so caching them based on their species would not make sense. So if the species cache object passed in is null, then species-based caching will not happen.

src/main/java/org/reactome/addlinks/fileprocessors/PharmacoDBFileProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+		// Now we need to create the map that goes from IUPHAR to PharmacoDB.
+		for (Entry<String, String> entry : iuphar2PubChem.entrySet())
+		{
+			String iuphar = entry.getKey();
+			String pubChem = entry.getValue();
+			String pharmacoDB = pubChem2Pharmacodb.get(pubChem);
+
+			// The regex "PDBC\d+" is because the sample file I got from PharmacoDB sometimes has none-identifiers in the PharmacoDB.uid field (such as "TRUE", "NA", etc...)
+			// Hopefully those issues will be resolved by the time they are publishing this file live.
+			if (pharmacoDB != null && !pharmacoDB.isEmpty() && !pharmacoDB.isBlank() && pharmacoDB.matches("PDBC\\d+"))
```
jweiser, line 95: `isBlank()` needs to be removed/replaced since this is after Java 8.

src/main/resources/basic-file-retrievers.xml
```diff
@@ -294,5 +294,23 @@
 				<constructor-arg name="retrieverName" value="retrievers/EnsemblBioMart"/>
 			</bean>
 		</entry>
+
+		<!-- NOTE: The files retrieved from PharmacoDB and IUPHAR are *both* required to create references for PharmacoDB -->
+		<entry key="PharmacoDB">
+			<bean class="org.reactome.addlinks.dataretrieval.FileRetriever" id="PharmacoDB">
+				<property name="dataURL" value="https://raw.githubusercontent.com/BHKLAB-Pachyderm/Annotations/master/drugs_with_ids.csv" />
```
jweiser, line 301: Does this need to be changed now that they're not in beta anymore?

src/main/resources/file-download-context.xml
```diff
@@ -71,6 +71,8 @@
 		<value>ComplexPortal</value>
 		<value>ComplexPortalSARS</value>
 		<value>ComplexPortalSARSCOV2</value>
+		<value>IUPHAR</value>
```
jweiser, line 74: You might need to do a search and replace for "IUPHAR" to "Guide To Pharmacology" in the project since it's their new name and the reference database in gk_central will be renamed very soon.

src/main/resources/basic-file-retrievers.xml
SolomonShorser-OICR: That _is_ the URL to their file. Previously, it was emailed to me and I placed it in the correct directory for the file processor to pick up.

src/main/resources/file-download-context.xml
SolomonShorser-OICR: Ugh... I was hoping to leave that for Release 77. :/

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
SolomonShorser-OICR: I think I spoke to someone (maybe you?) waaaaaay back when I first wrote this and it seemed like an allowable condition, at the time. You can raise it, if you want.

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
jweiser: True, but I think the cast is worth it because as a new reader of the code, I may have no idea what type species is if it is `Object` since that can hold any Java object!  For the sake of a cast (since getAttributeValue for species is guaranteed to return a GKInstance), I think it would help someone new to the code so they know its type.

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
jweiser: I can think of cases where it may be okay, but it's probably best to confirm with the curators.

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
jweiser: Okay, just wanted to check the method shouldn't do something if it receives a null value for that parameter.  Is it worth checking that upfront in the method so you don't have to do anything else if the method receives a null?

src/main/resources/basic-file-retrievers.xml
jweiser: Okay, I just wanted to double-check because a URL in raw.githubusercontent looked "beta-ish"

src/main/resources/file-download-context.xml
jweiser: Does the program use this name to try to grab the reference database instance?  If not, you could leave it.  If so, it may still work because "IUPHAR" will be one of the values in the "name" attribute but it won't be the first one or the "displayName"

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
SolomonShorser-OICR: It might be worth moving the check to the beginning. This code has been refactored many times - at one time I think this code was a part of a larger method that did other stuff even when that specific parameter was null.

src/main/resources/basic-file-retrievers.xml
SolomonShorser-OICR: I know. But... that's the URL Gangesh gave me... :roll_eyes:

src/main/resources/file-download-context.xml
SolomonShorser-OICR: IUPHAR is not in reference-databases.xml, so it doesn't really matter what name is used in the Spring context files. I think changing the Spring identifier "IUPHAR" to "GtP" (or whatever) can wait until later (i.e. when I'm less busy).

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
SolomonShorser-OICR: This is old code so I'm guessing that the exception was getting caught and the stack trace was _not_ getting printed, for some reason.

src/main/java/org/reactome/addlinks/db/ReferenceObjectCache.java
SolomonShorser-OICR: No, you need to explicitly request something that will be compatible with `Collections.synchronizedList`. If you just use `Arrays.asList` you get this stack trace:
```
java.lang.UnsupportedOperationException
	at java.base/java.util.AbstractList.add(AbstractList.java:153)
	at java.base/java.util.AbstractList.add(AbstractList.java:111)
	at java.base/java.util.Collections$SynchronizedCollection.add(Collections.java:2040)
	at org.reactome.addlinks.db.ReferenceObjectCache.addObjectToIdentifierCache(ReferenceObjectCache.java:166)
...
```

jweiser: Okay, thanks for the explanation and info!

jweiser (approved):
I think everything I've mentioned has been addressed.  Thanks!

PR #158, opened 2021-03-12
R76 tweaks

SolomonShorser-OICR:
Quick fixes to ensure that KEGG "vg" (virus) and ag ("addendum") database prefixes are in the KEGG species cache.

When running the UniProt-mapped-to-KEGG reference creation process, it failed with "vg:"-prefixed identifiers because "vg" was not in the KEGG species cached. I assume that this error was not noticed before because we did not get "vg"-prefixed identifiers mapped by UniProt in the past.

### DIFF
--- a/src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
+++ b/src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
@@ -40,7 +40,7 @@ public class KEGGReferenceDatabaseGenerator
 		KEGGReferenceDatabaseGenerator.dbCreator = creator;
 	}
 
-	private static Long createReferenceDatabase(String dbName, String speciesName, String speciesURL, ReferenceObjectCache objectCache) throws Exception
+	private static synchronized Long createReferenceDatabase(String dbName, String speciesName, String speciesURL, ReferenceObjectCache objectCache) throws Exception
 	{
 		if (objectCache.getRefDbNamesToIds().keySet().contains(dbName))
 		{
@@ -69,7 +69,7 @@ public class KEGGReferenceDatabaseGenerator
 	 * Generate new species-specific KEGG reference databases. The new database will be named in the form "KEGG Gene (${SPECIES_NAME})".
 	 * @param objectCache - The object cache. The list of species in this cache will be what the new reference databases are based on.
 	 */
-	public static void generateSpeciesSpecificReferenceDatabases(ReferenceObjectCache objectCache)
+	public static synchronized void generateSpeciesSpecificReferenceDatabases(ReferenceObjectCache objectCache)
 	{
 		for (String speciesName : objectCache.getSpeciesNamesToIds().keySet())
 		{
@@ -139,7 +139,7 @@ public class KEGGReferenceDatabaseGenerator
 	 * @param keggSpeciesCode KEGG species code.
 	 * @return The DBID of the KEGG ReferenceDatabase object, if it exists. If no ReferenceDatabase can be found (or too many matching ReferenceDatabases), then NULL will be returned.
 	 */
-	public static Long getKeggReferenceDatabase(String keggSpeciesCode)
+	public static synchronized Long getKeggReferenceDatabase(String keggSpeciesCode)
 	{
 		Long dbId = null;
 		if (KEGGReferenceDatabaseGenerator.keggCodesToRefDBMap.containsKey(keggSpeciesCode))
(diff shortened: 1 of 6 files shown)


src/main/resources/file-download-context.xml
```diff
@@ -31,20 +31,21 @@
...
 		<constructor-arg index="1" type="boolean" name="lazyLoad" value="${lazyLoadCache}" />
 	</bean>
-	
+
 
 	<util:list id="fileRetrieverFilter" value-type="java.lang.String">
-		<value>UniProtToKEGG</value>
-		<value>KEGGRetriever</value>
+		<value>COSMIC</value>
+<!--		<value>IntEnz</value>
```
jweiser, line 44: Is the commenting out just for one run?  If so, should it be removed from the PR?

src/main/java/org/reactome/addlinks/kegg/KEGGSpeciesCache.java
```diff
@@ -103,9 +104,19 @@ private KEGGSpeciesCache()
 						// or summary/group headings.
 						logger.trace("Line/pattern mismatch: {}",line);
 					}
-
 				}
-				logger.info("{} keys added to the KEGG species map.",KEGGSpeciesCache.speciesMap.keySet().size());
+				// Add vg for Virus, and ag for Addendum - they are not in the KEGG species list, but are still valid prefixes.
+				codesToSpecies.put("vg", "Virus");
+				codesToSpecies.put("ag", "Addendum");
+				HashMap<String, String> virusMap = new HashMap<>();
```
jweiser, line 111: Better to use Map<String, String>?

src/main/java/org/reactome/addlinks/kegg/KEGGSpeciesCache.java
```diff
@@ -103,9 +104,19 @@ private KEGGSpeciesCache()
...
+				virusMap.put(KEGG_CODE, "vg");
+				virusMap.put(COMMON_NAME, "");
+				speciesMap.put("Virus", Arrays.asList(virusMap));
+				HashMap<String, String> addendumMap = new HashMap<>();
```
jweiser, line 115: Same as https://github.com/reactome/add-links/pull/158/files#r599069739

src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
```diff
@@ -245,7 +245,7 @@ public static synchronized String createReferenceDatabaseFromKEGGData(String keg
 		String speciesURL = KEGG_URL.replace("###SP3###", keggPrefix + ":");
 		try
 		{
-			Long dbId = new Long(createReferenceDatabase(newDBName, keggSpeciesName, speciesURL, objectCache));
+			Long dbId = createReferenceDatabase(newDBName, keggSpeciesName, speciesURL, objectCache);
```
jweiser, line 248: Not a huge deal, but generally using primitives (i.e. `long` vs `Long`) avoid pitfalls of comparing using `==` if that is ever added.

SolomonShorser-OICR: Yes... the Reactome API uses `Long`, though, so that is probably why it was done. I'll have a look.

SolomonShorser-OICR: ...it gets passed to `adaptor.fetchInstance`, I guess I wanted to match the exact type that the method required.

jweiser: Okay, it's not critical.  You can use `long` if you want.  Java will automatically unbox it.

jweiser (approved):

src/main/java/org/reactome/addlinks/kegg/KEGGReferenceDatabaseGenerator.java
SolomonShorser-OICR: If we ever get to migrate to Java 16+, there are changes coming for primitive wrappers: https://openjdk.java.net/jeps/390

jweiser: Okay, looks like additional reasons to favour primitives instead of classes when possible.

SolomonShorser-OICR: Yes. If only we could rewrite the underlying Reactome API...

jweiser: We might be able to make some rewrites for the project in the future.

PR #162, opened 2021-05-04
Feature/add pharos

SolomonShorser-OICR:
Pharos data retrieval.


src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,159 @@
...
+				// But there just isn't time before Release 77 for that sort of elegant solution, so yes, there *is* some duplication of effort in this
+				// doHttpDownload method and the method it overrode.
+				HttpPost post = new HttpPost(PharosDataRetriever.pharosEndpoint);
+
+
+				// We will need to keep making posts until we have them all. Requests will get 100 items at a time.
+				RequestConfig config = RequestConfig.copy(RequestConfig.DEFAULT)
+											.setConnectTimeout(1000 * (int)this.timeout.getSeconds())
+											.setSocketTimeout(1000 * (int)this.timeout.getSeconds())
+											.setConnectionRequestTimeout(1000 * (int)this.timeout.getSeconds())
```
jweiser, line 68: Maybe a temp variable for this so it can be reused: `int millisecondsTimeOut = 1000 * (int) this.timeout.getSeconds();`

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,159 @@
...
+										.replace(PharosDataRetriever.TOP_AMOUNT_TOKEN, Integer.toString(PharosDataRetriever.batchSize))
+										.replace(PharosDataRetriever.SKIP_AMOUNT_TOKEN, Integer.toString(numRequests * PharosDataRetriever.batchSize));
+				StringEntity stringEntity = new StringEntity(updatedQuery);
+				stringEntity.setContentType(PharosDataRetriever.APPLICATION_JSON);
+				post.setEntity(stringEntity);
+				post.setHeader(HttpHeaders.ACCEPT, PharosDataRetriever.APPLICATION_JSON);
+				post.setHeader(HttpHeaders.CONTENT_TYPE, PharosDataRetriever.APPLICATION_JSON);
+				int retries = this.numRetries;
+				boolean done = retries + 1 <= 0;
+				while (!done)
```
jweiser, line 84: I find the `retries + 1 <= 0` a bit confusing.  Could the condition be `retries < 0` since the condition will evaluate to the same thing?

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,159 @@
...
+					}
+					catch (ConnectTimeoutException e)
+					{
+						// we will only be retrying the connection timeouts, defined as the time required to establish a connection.
+						// we will not handle socket timeouts (inactivity that occurs after the connection has been established).
+						// we will not handle connection manager timeouts (time waiting for connection manager or connection pool).
+						e.printStackTrace();
+						logger.info("Failed due to ConnectTimeout, but will retry {} more time(s).", retries);
+						retries--;
+						done = retries + 1 <= 0;
```
jweiser, line 123: Same as here: https://github.com/reactome/add-links/pull/162/files#r626877164

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,159 @@
...
+					catch (HttpHostConnectException e)
+					{
+						logger.error("Could not connect to host {} !",post.getURI().getHost());
+						e.printStackTrace();
+						throw e;
+					}
+					catch (IOException e) {
+						logger.error("Exception caught: {}",e.getMessage());
+						throw e;
+					}
```
jweiser, line 138: Could the stack trace print on line 132 be part of the logger message on line 131?

`logger.error("Could not connect to host {}: {} !", post.getURI().getHost(), e.getMessage())`;

This is done on line 136.  It should be okay not to line 132 because the exception gets rethrown.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,159 @@
...
+				}
+			}
+		}
+	}
+
+	/**
+	 * Implementors must return a specific query to be executed by this class' doHttpDownload
+	 * @return
```
jweiser, line 146: Could the JavaDoc description be added on what is returned for completeness?

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,159 @@
...
+	protected abstract String getQuery();
+
+	/**
+	 * Implementors must process an array of JSON objects.
+	 * @param fw A FileWriter - this can be used to write data to the output file.
+	 * @param jsonObj - the JSON object to process.
+	 * @return The number of items processed.
+	 * @throws IOException
+	 */
+	protected abstract int processJSONArray(FileWriter fw, JSONObject jsonObj) throws IOException;
```
jweiser, line 157: Could the FileWriter parameter be `fileWriter` instead of `fw` just so it's clear what it is when in the body of the code without the type declaration beside it?

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,159 @@
...
+	 * @return
+	 */
+	protected abstract String getQuery();
+
+	/**
+	 * Implementors must process an array of JSON objects.
+	 * @param fw A FileWriter - this can be used to write data to the output file.
+	 * @param jsonObj - the JSON object to process.
+	 * @return The number of items processed.
+	 * @throws IOException
```
jweiser, line 155: Could the JavaDoc description be added here for completeness on when the exception is expected to be thrown?

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
```diff
@@ -0,0 +1,78 @@
...
+			+ "      synonyms {"
+			+ "        name"
+			+ "        value"
+			+ "      }"
+			+ "    }"
+			+ "  }"
+			+ "}\"}";
+
+	@Override
+	protected int processJSONArray(FileWriter fw, JSONObject jsonObj) throws IOException
```
jweiser, line 31: Same as here: https://github.com/reactome/add-links/pull/162/files#r626881055

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
```diff
@@ -0,0 +1,78 @@
...
+	{
+		try
+		{
+			JSONArray ligands = (((JSONObject)((JSONObject) jsonObj.get("data")).get("ligands")).getJSONArray("ligands"));
+			for (int i = 0; i < ligands.length(); i++)
+			{
+				JSONObject ligand = ((JSONObject)ligands.get(i));
+				Boolean isDrug = ligand.getBoolean("isdrug");
+				// Check that the ligand has "isdrug==true" and then get the GtP synonym's value (synonym's "name" will be "Guide to Pharmacology").
+				if (isDrug.booleanValue())
```
jweiser, line 41: Could line 39 be `boolean isDrug = ligand.getBoolean("isdrug")`?  Java will automatically unbox the object Boolean to a primitive and that way line 41 could just be `if(isDrug)`.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
```diff
@@ -0,0 +1,78 @@
...
+
+				}
+			}
+			return ligands.length();
+		}
+		catch (JSONException e)
+		{
+			logger.error("Error processing JSON! JSON is: {}", jsonObj.toString());
+			logger.error("Exception is: ", e);
+			return 0;
```
jweiser, line 69: Is this the number of ligands processed?  If so, it's possible some of the ligands could be processed and written to the FileWriter before an exception is caught.  Maybe a `ligandsProcessedCount` might be better to be returned.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
```diff
@@ -0,0 +1,78 @@
...
+					// Now we process the synonyms...
+					JSONArray synonyms = ligand.getJSONArray("synonyms");
+					for (int j = 0; j < synonyms.length(); j++)
+					{
+						JSONObject synonym = (JSONObject)synonyms.get(j);
+						if (synonym.getString("name").equals("Guide to Pharmacology"))
+						{
+							gtpIdentifier = synonym.getString("value");
+						}
+					}
```
jweiser, line 55: Maybe a while loop is better here because the loop will keep executing even after finding the `gtpIdentifier`.

```
int synonymIndex = 0;
while (gtpIdentifier == null && synonymIndex < synonyms.length()) {
    //current code
    synonymIndex++;
}

File: src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosTargetsDataRetriever.java [hunk 12]
```diff
@@ -0,0 +1,43 @@
...
+			+ "            uniprot"
+			+ "        }"
+			+ "    }"
+			+ "}\" }";
+
+
+	/**
+	 * Implements the logic to process JSON from Pharos.
+	 */
+	protected int processJSONArray(FileWriter fw, JSONObject jsonObj) throws IOException
```
jweiser, line 26: Same as here: https://github.com/reactome/add-links/pull/162/files#r626881913

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosTargetsDataRetriever.java
```diff
@@ -0,0 +1,43 @@
...
+	{
+		JSONArray targets = (((JSONObject)((JSONObject) jsonObj.get("data")).get("targets")).getJSONArray("targets"));
+		for (int i = 0; i < targets.length(); i++)
+		{
+			fw.write(((JSONObject)targets.get(i)).get("uniprot") + "\n");
```
jweiser, line 31: Is it worth checking to see if the JSONObject has a "uniprot" key or it's result is not null?  Otherwise a null value will be written to the FileWriter.

SolomonShorser-OICR: The query sent to the server specifies "uniprot" so there should be a "uniprot" key.

src/test/java/org/reactome/addlinks/test/PharosDataRetrieverIT.java
```diff
@@ -0,0 +1,67 @@
...
+		String pathToPharosLigandsFile = "/tmp/pharos_ligand_data.txt";
+		PharosDataRetriever retriever = new PharosLigandDataRetriever();
+		testRetriever(pathToPharosLigandsFile, retriever);
+	}
+
+	/**
+	 * Tests a Pharos retriever.
+	 * @param pathToPharosFile Path to output file.
+	 * @param retriever A retriever object to test.
+	 * @throws URISyntaxException
```
jweiser, line 46: Could the JavaDoc description be added for completeness on when the URISyntaxException will be thrown?

src/test/java/org/reactome/addlinks/test/PharosDataRetrieverIT.java
```diff
@@ -0,0 +1,67 @@
...
+			retriever.fetchData();
+			assertTrue(Files.exists(Paths.get(pathToPharosFile), LinkOption.NOFOLLOW_LINKS));
+			assertTrue(Files.size(Paths.get(pathToPharosFile)) > 0);
+		}
+		catch (Exception e)
+		{
+			e.printStackTrace();
+			fail();
+		}
+	}
```
jweiser, line 65: Does the downloaded data file need to be deleted after the test so this test be run more than once automatically?

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
SolomonShorser-OICR: I think if this happens it happens _before_ any real data gets processed, or not at all. But if this exception is thrown, the process should be rerun since not all (or any) data was properly processed.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
SolomonShorser-OICR: Yes, a while-loop is more elegant, but honestly, I was just in a rush to get _something_ working.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
SolomonShorser-OICR: Hmm I can try that.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
jweiser: My concern is more about that the loop keeps running after it's found what it searched for.  It's not a huge deal.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosTargetsDataRetriever.java
jweiser: Okay, I was just thinking about the possibility if it's somehow is not a key.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
SolomonShorser-OICR: Yes, I understand the concern. Extra synonyms will be examined, and milliseconds will be wasted. But Robin wants this run as a part of Release 77. So some changes may be deferred.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
jweiser: Okay

src/test/java/org/reactome/addlinks/test/PharosDataRetrieverIT.java
SolomonShorser-OICR: No because of `retriever.setMaxAge(Duration.ofSeconds(0));`. I'd prefer to not delete them because it's useful to keep them around for manual inspection & debugging.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
jweiser: Okay, that's fine

src/test/java/org/reactome/addlinks/test/PharosDataRetrieverIT.java
jweiser: Okay, as long as it doesn't interfere with re-running the test, then I'm okay with it.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
SolomonShorser-OICR: Just to clarify: your code suggestion will _not_ print the stack trace, it will _only_ print the message. I'm fine with that, if that's what you meant.

jweiser: Yeah, I was thinking to print the message to be in sync with the way the next exception block logs the exception.  You could pass the entire exception object to the logger as the last argument instead of the message if you want so the entire stack trace is logged.

I also thought since the exception is rethrown another part of the code may print the stack trace so it may be okay to take out the stack trace to screen here.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosTargetsDataRetriever.java
SolomonShorser-OICR: Here's what the data looks like:
```
{
  "data": {
    "targets": {
      "count": 20412,
      "targets": [
        {
          "uniprot": "A8MVA2"
        },
        {
          "uniprot": "A8MX34"
        },
...
```
If there is no "uniprot" then something has gone wrong.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
SolomonShorser-OICR: The statement from the "next exception block" (`logger.error("Exception caught: {}",e.getMessage());`) does not log the entire exception. Only the message.

I changed the code to only log the messages, I thought that's what you wanted.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
```diff
@@ -27,35 +27,107 @@
...
+					while (!done)
 					{
 						JSONObject synonym = (JSONObject)synonyms.get(j);
 						if (synonym.getString("name").equals("Guide to Pharmacology"))
 						{
 							gtpIdentifier = synonym.getString("value");
+							done = true;
 						}
+						j++;
+						done = j >= synonyms.length();
```
jweiser: I think this will still run for the full loop even after finding the gtpIdentifier because done gets reassigned on line 126 even if it was set to true on line 123 if the counter "j" is less than the synonym length.

It think it would be better to have the whole condition in the while loop itself instead of using `done` as an intermediate because you can say `while(gtpIdentifier == null || j < synonyms.length())` in one place.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
```diff
@@ -64,9 +136,11 @@ protected int processJSONArray(FileWriter fw, JSONObject jsonObj) throws IOExcep
 		}
 		catch (JSONException e)
 		{
-			logger.error("Error processing JSON! JSON is: {}", jsonObj.toString());
+			logger.error("Error processing JSON! JSON source is: {}", jsonObj.toString());
 			logger.error("Exception is: ", e);
-			return 0;
+			// Still need to return something so return the number of items examined.
+			// But if a JSONException occurs, this will very likely be 0.
+			return i;
```
jweiser, line 143: Thanks for making this change!

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosTargetsDataRetriever.java
```diff
@@ -19,16 +19,39 @@
...
-			fw.write(((JSONObject)targets.get(i)).get("uniprot") + "\n");
+			if (((JSONObject)targets.get(i)).has("uniprot"))
+			{
+				writer.write(((JSONObject)targets.get(i)).get("uniprot") + "\n");
+			}
+			else
+			{
+				logger.error("The \"uniprot\" key was not present, but it should be (in fact, it should be the ONLY key at this depth in the document). Check your query!");
+				// Something's wrong with the data, so gracefully exit the loop.
+				i = targets.length() + 1;
```
jweiser, line 53: Could this be a `break;` instead?  I didn't understand immediately on reading that the loop is exiting by modifying the value of i to be one more than `targets.length()`.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosTargetsDataRetriever.java
jweiser: Okay I see.  I see you added a logger statement to handle it.  Thanks!

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
jweiser: I'm okay either way - I just was mentioning another possibility.  I'm fine with the way the code is now.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosLigandDataRetriever.java
SolomonShorser-OICR: Good catch!

src/test/java/org/reactome/addlinks/test/PharosDataRetrieverIT.java
jweiser: I saw you added all the JavaDoc changes which look good!  I think this one is the last one.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosTargetsDataRetriever.java
SolomonShorser-OICR: I don't think this would be an appropriate use of `break`.

jweiser: Is there a reason why?  `break;` is meant to stop a loop (or switch statement).  I think it is important because modifying the counter within the for loop usually isn't done outside of the for statement itself.

SolomonShorser-OICR: Honestly, I have a hard time coming up with _appropriate_ uses of `break`.

I think that using `break` is a bad practice in general. There's almost always a way to clearly define the exit conditions from a loop. `break` leads to hidden traps and bad habits. I've seen code where break and continue were used in horrible ways to make code that was almost completely incomprehensible.

I don't understand what's wrong with explicitly violating the loop invariant to exit the loop early.

SolomonShorser-OICR: Well... switch-case statements use `break`, but that's the only place I use them, and that's because they are designed that way.

jweiser: I am just very uncomfortable with modifying the loop condition in a for loop.  I've never seen it done.  I really do think this is an appropriate use for break, here.

If you'd prefer not to use break, I think this would be better as a while loop so that you can include the counter and the condition of error to stop.

SolomonShorser-OICR: Modifying a variable to break a for-loop's condition is not common, but I have seen it done and I still think it's more appropriate than a `break`. I'll probably rewrite this as a while-loop since I think that's the only thing we'll agree on.

SolomonShorser-OICR: I've changed the way this possible error gets handled. Please have a look.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -154,7 +166,8 @@ protected void doHttpDownload(Path path, HttpClientContext context) throws Excep
 	 * @param jsonObj - the JSON object to process.
 	 * @return The number of items processed.
 	 * @throws IOException Thrown when there is a problem writing the file.
+	 * @throws PharosDataException
```
jweiser, line 191: Could a JavaDoc description be added here?

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosTargetsDataRetriever.java
```diff
@@ -36,24 +37,34 @@
 
 	/**
 	 * Implements the logic to process JSON from Pharos.
+	 * @throws PharosDataException
```
jweiser, line 40: Same as https://github.com/reactome/add-links/pull/162/files/aa16cfa69985fc2c2f1ca117d718abe0c9d8622a..858f0cb037e77d6cc30cde26f76606b83dd68cee#r628469233

src/test/java/org/reactome/addlinks/test/PharosDataRetrieverIT.java
```diff
@@ -20,9 +49,24 @@
  * @author sshorser
  *
  */
+//@RunWith(PowerMockRunner.class)
+//@PrepareForTest({ EntityUtils.class, HttpEntity.class })
+//@PowerMockIgnore({"jdk.internal.reflect.*", "jdk.xml.internal.*", "javax.xml.*", "org.apache.logging.*", "javax.management.*","javax.net.ssl.*", "org.xml.*"})
+//@PrepareForTest({ org.apache.http.impl.client.HttpClients.class, PharosDataRetriever.class, PharosTargetsDataRetriever.class})
```
jweiser: Can this be removed along with imports for PowerMock if it's not needed?

jweiser (commented):
Looks good!  Just noticed a couple of JavaDoc tags with descriptions missing, but I'm fine to approve the PR after they're added.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosTargetsDataRetriever.java
jweiser: Looks good!  Thanks for making the change!

src/test/java/org/reactome/addlinks/test/PharosDataRetrieverIT.java
SolomonShorser-OICR: I think you were looking at an earlier commit - it should have been cleaned up in https://github.com/reactome/add-links/pull/162/commits/5e286e05583282507d4a99fcf95137b21c0dfe7b

jweiser: Yeah, I think I was - sorry about that!

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,195 @@
...
+	// curl 'https://ncatsidg-dev.appspot.com/graphql' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: https://ncatsidg-dev.appspot.com' --data-binary '{"query":"query {  targets {    targets(top:100) {      uniprot    }  }}\n"}' --compressed
+	private static int batchSize = 1000;
+
+	@Override
+	protected void doHttpDownload(Path path, HttpClientContext context) throws Exception, HttpHostConnectException, IOException
```
jweiser, line 51: Does it make sense to include `PharosDataException` in the method signature that is thrown from this method now?

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,195 @@
...
+						logger.error("IOException caught: {}", e.getMessage());
+						throw e;
+					}
+				}
+			}
+		}
+	}
+
+	/**
+	 * Report on a not-OK (non-200) HTTP response.
```
jweiser, line 130: Maybe mention that the method does not report anything if the `statusCode` passed happens to be 200?

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -0,0 +1,195 @@
...
+	 * @param response - Response object.
+	 * @param statusCode - status code.
+	 */
+	private void reportNotOKResponse(StatusLine statusLine, int statusCode)
```
jweiser, line 134: The first `@param` looks like it needs updating to be `statusLine`

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
SolomonShorser-OICR: That was masked by the `throws Exception`. And yes, I would like to improve exception-handling in our applications in general, but no, this is not the time to do it.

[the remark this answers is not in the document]
SolomonShorser-OICR: That hasn't been done for any other similar situation in this application, but I'll add it in anyway.

jweiser: Yeah, but since HttpHostConnectException and IOException are included, it would make sense to include PharosDataException or all can be omitted if just Exception is kept.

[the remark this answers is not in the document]
jweiser: Okay, just mentioning but not a problem either way.  I wasn't sure if you usually add JavaDocs to protected methods or not.

jweiser (approved):

PR #163, opened 2021-05-07
Feature/pharos reference creation

SolomonShorser-OICR:
Add new file processors for Pharos
Update configuration to run Pharos-related code in whole AddLinks process.


src/main/java/org/reactome/addlinks/fileprocessors/pharos/PharosLigandsFileProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+			}
+		}
+		catch (FileNotFoundException e)
+		{
+			logger.error("File {} could not be found, file processor has failed.", this.pathToFile.toString());
+		}
+		catch (IOException e)
+		{
+			logger.error("Error with file! File processing has probably failed.", e);
+		}
```
jweiser, line 47: For consistency, can the exception object `e` be logged on line 42?

SolomonShorser-OICR: I didn't like the Exception object because I felt that a `FileNotFoundException`, combined with the name of the file on the path that it was _supposed_ to be found on was already pretty clear. I logged the `IOException` object because there could be more than one line of code that triggers it, and there could be more than one cause for it, so I felt that logging the exception object was warranted in this case.

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
```diff
@@ -49,6 +49,16 @@
...
 	private static int batchSize = 1000;
 
+	public PharosDataRetriever(String retrieverName)
+	{
+		super(retrieverName);
+	}
+
+	public PharosDataRetriever()
+	{
+		super(null);
```
jweiser, line 59: Just wanted your input - is it worth passing an empty String instead of null for no retriever name so there are no exceptions due to null values?

src/main/java/org/reactome/addlinks/fileprocessors/pharos/PharosLigandsFileProcessor.java
jweiser: Okay, would the logged error indicate what class triggered it though so it's clear where in code the file that wasn't findable is being used?

src/main/java/org/reactome/addlinks/dataretrieval/pharos/PharosDataRetriever.java
SolomonShorser-OICR: `CustomLoggable` can handle this, but in the future, I believe that empty strings will work as well.

src/main/java/org/reactome/addlinks/fileprocessors/pharos/PharosTargetsFileProcessor.java
```diff
@@ -0,0 +1,54 @@
...
+	public Map<String, String> getIdMappingsFromFile()
+	{
+		Map<String, String> mapping = new HashMap<>();
+
+		try(BufferedReader reader = new BufferedReader(new FileReader(this.pathToFile.toFile())))
+		{
+			String line = reader.readLine();
+			while (line != null)
+			{
+				mapping.put(line, line);
```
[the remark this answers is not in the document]
jweiser: Okay, I kind of figured that from the comment but I just wanted to make sure.

src/main/java/org/reactome/addlinks/fileprocessors/pharos/PharosLigandsFileProcessor.java
SolomonShorser-OICR: Passing the exception to logger.error should include the exception's stack trace in the log message.

jweiser: Okay, sounds good.

jweiser (approved):

PR #171, opened 2021-06-22
Handle deprecated identifiers

### LINKED ISSUE #170: Check accessURL from identifiers.org for "deprecated"
identifiers.org may deactivate/deprecate a resource identifier. If this happens, the web request we send will return valid JSON and a 200 response code, BUT the accessURL will be something like `https://registry.identifiers.org/deprecation/resources/MIR:00100113/{$id}` so _don't_ use that accessURL if that's what comes back. Also, write an error message so the users will know about it.

SolomonShorser-OICR:
Fixes #170

### DIFF
--- a/src/main/java/org/reactome/addlinks/IdentifiersDotOrgUtil.java
+++ b/src/main/java/org/reactome/addlinks/IdentifiersDotOrgUtil.java
@@ -19,7 +19,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 /**
- * Utility methods for querying the service endpoints at http://identifiers.org 
+ * Utility methods for querying the service endpoints at http://identifiers.org
  * @author sshorser
  *
  */
@@ -29,7 +29,7 @@ public class IdentifiersDotOrgUtil
 	private static final String ACCESS_URL_TOKEN = "urlPattern";
 	private static final String RESOURCE_IDENTIFIER_ENDPOINT = "https://registry.api.identifiers.org/restApi/resources/search/findByMirId?mirId=";
 	private static final Logger logger = LogManager.getLogger();
-	
+
 	/**
 	 * Gets the accessURL for a resource identifier from identifiers.org. Resource identifiers usually have a prefix and identifier string.
 	 * For example, "MIR:00100374" refers to the resource "HAPMAP". If you request <code>https://identifiers.org/rest/resources/MIR:00100374</code>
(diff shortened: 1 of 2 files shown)


src/main/java/org/reactome/addlinks/IdentifiersDotOrgUtil.java
```diff
@@ -69,11 +70,27 @@ public static String getAccessUrlForResource(String resourceIdentifier)
...
+						// check that the identifier is not deprecated. If identifiers.org
+						// deprecates an identifier, the AccessURL in the response
+						// will look something like this: https://registry.identifiers.org/deprecation/resources/MIR:00100113/{$id}
+						// (this example uses the deactivated/deprecated Rhea identifier that caused a problem
+						// during Releaes 77)
+						// But the HTTP Response code is 200 so we can't just blindly replace our Access URL with the
+						// one from identifiers.org.
+						if (responseAccessURL.contains("registry.identifiers.org/deprecation"))
+						{
+							logger.error("De-activated/deprecated identifier detected for resource identifier {}\nYou should check on identifiers.org to find a new resource identifier for your resource.", resourceIdentifier);
```
jweiser, line 84: Is this serious enough to throw an exception or is logging sufficient?

src/main/java/org/reactome/addlinks/IdentifiersDotOrgUtil.java
```diff
@@ -48,14 +48,15 @@
 	 * @param resourceIdentifier - The identifier of a resource, as per identifiers.org. If you are uncertain what the resourceIdentifier is,
 	 * you can search the identifiers.org website, or see a complete list of resources here:
 	 * <a href="https://identifiers.org/rest/resources">https://identifiers.org/rest/resources</a>
-	 * @return The accessURL of the resource identified by <code>resourceIdentifier</code>.
+	 * @return The accessURL of the resource identified by <code>resourceIdentifier</code>. In cases where identifiers.org does not return an accessURL, or where the
+	 * identifier is deactivated/deprecated (according to identifiers.org), NULL will be returned.
```
jweiser, line 52: Does the caller or callers of this method need to be updated to check for a NULL value?

SolomonShorser-OICR: No. This is not new, only the comment that explicitly explains the NULL value is new.

src/main/java/org/reactome/addlinks/IdentifiersDotOrgUtil.java
SolomonShorser-OICR: :shrug: 
Currently, an error is logged and the accessURL in the reference-databases.xml config file will be used. 

If an exception is thrown, it should probably interrupt the program and cause it to fail. This means that AddLinks will _not_ be able to run if there is an out-of-date resource identifier. A way to get around that is to simply omit the resource identifier (not all databases have resource identifiers in identifiers.org so they are optional). 

We can ask Robin how he wants this handled.

src/main/java/org/reactome/addlinks/IdentifiersDotOrgUtil.java
jweiser: Okay, sounds good - I just wanted to check.

src/main/java/org/reactome/addlinks/IdentifiersDotOrgUtil.java
jweiser: Okay - I'm not advocating for either of these.  I just wanted to ask because I didn't know.

jweiser (approved):

PR #172, opened 2021-06-23
R77 tweaks

SolomonShorser-OICR:
Some fixes that came up during Release 77

### DIFF
--- a/src/main/resources/application-context.xml
+++ b/src/main/resources/application-context.xml
@@ -40,7 +40,7 @@
 		<value>GuideToPharmacologyLigands</value>
 		<value>GuideToPharmacologyTargets</value>
 		<value>PharmacoDB</value>
-		<value>IUPHAR</value>
+		<!-- <value>IUPHAR</value> --><!-- Renamed to GuideToPharmacologyLigands -->
 		<value>ComplexPortal</value>
 		<value>ComplexPortalSARS</value>
 		<value>ComplexPortalSARSCOV2</value>
@@ -192,6 +192,7 @@
 		<value>Guide to Pharmacology - Targets</value>
 		<value>Guide to Pharmacology - Ligands</value>
 		<value>Target Pathogen</value>
+		<value>PharmacoDB</value>
 		<value>IntEnz</value>
 		<value>ENSEMBL</value>
 		<value>PDB</value>
--- a/src/main/resources/application-context.xml
+++ b/src/main/resources/application-context.xml
@@ -40,7 +40,6 @@
 		<value>GuideToPharmacologyLigands</value>
 		<value>GuideToPharmacologyTargets</value>
 		<value>PharmacoDB</value>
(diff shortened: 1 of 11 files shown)


src/main/resources/application-context.xml
```diff
@@ -40,7 +40,7 @@
 		<value>GuideToPharmacologyLigands</value>
 		<value>GuideToPharmacologyTargets</value>
 		<value>PharmacoDB</value>
-		<value>IUPHAR</value>
+		<!-- <value>IUPHAR</value> --><!-- Renamed to GuideToPharmacologyLigands -->
```
jweiser, line 43: Should this remain commented out or does it need to be replaced with the new name?

SolomonShorser-OICR: `GuideToPharmacologyLigands` is already used a few lines above. That note was more of an explanation to myself (or anyone else who remembers IUPHAR). I can remove this line, if you would like.

jweiser: Oh okay - I didn't notice the value above.  Maybe it could be moved to just above "GuideToPharmacologyLigands" and the comment could be "IUPHAR renamed to GuideToPharmacologyLigands below" so the comment leads right to the replacement.

SolomonShorser-OICR: Maybe I'll just delete it...

src/main/resources/file-download-context.xml
```diff
@@ -73,13 +77,9 @@
 		<value>HmdbProteins</value>
 		<value>HmdbMolecules</value>
 		<value>RHEAIdentifierToReactionlikeEvent</value>
-		<value>ComplexPortal</value>
-		<value>ComplexPortalSARS</value>
-		<value>ComplexPortalSARSCOV2</value>
-		<value>IUPHAR</value>
+		<!-- <value>IUPHAR</value> --><!-- Renamed to GuideToPharmacologyLigands -->
```
jweiser, line 80: Same as here: https://github.com/reactome/add-links/pull/172/files?diff=unified&w=1#r657275281.  Moving it to just above "GuideToPharmacologyLigands" so the replacement is immediately after it.

src/main/resources/application-context.xml
jweiser: Okay, if you prefer.  You could add a comment after "GuideToPharmacologyLigands" to say "replaces IUPHAR", if you want to keep that note?

SolomonShorser-OICR: Already deleted.

jweiser: Okay, did you want to add the note to the "GuideToPharmacologyLigands" value to have a comment that says "replace IUPHAR" or no?

SolomonShorser-OICR: No.

jweiser: Okay

jweiser (approved):

